PR #69, opened 2024-05-16
Replace consequent `write` and `newLine` calls on CodeWriter with `writeLine`

jvmusin:
**Subsystem**
KSP plugin.

**Problem Description**
The code in `RPCClientServiceGenerator` would be easier to follow if it had no duplicated `write` and `newLine` methods which always go together.

**Solution**
Introduce `fun CodeWriter.writeLine(str: String)` which does writing given text and a newline character.
Since all the `write` method usages are followed by the `newLine` method call, the code becomes easier to read.
As a bonus, the `write` method is not used anymore and can be removed.

The changes in this commit do not change the file being written except for the final newline character.
Previously, generated files did not contain a newline character, and now they will.

### DIFF
--- a/ksp-plugin/src/main/kotlin/kotlinx/rpc/codegen/RPCClientServiceGenerator.kt
+++ b/ksp-plugin/src/main/kotlin/kotlinx/rpc/codegen/RPCClientServiceGenerator.kt
@@ -36,39 +36,29 @@ class RPCClientServiceGenerator(private val codegen: CodeGenerator) {
         writer: CodeWriter,
         service: RPCServiceDeclaration,
     ) {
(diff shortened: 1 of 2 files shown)


ksp-plugin/src/main/kotlin/kotlinx/rpc/codegen/RPCClientServiceGenerator.kt
```diff
@@ -79,27 +69,20 @@ class RPCClientServiceGenerator(private val codegen: CodeGenerator) {
             it.toCode(service.fullName, nested)
         }
         generateProviders(writer.nested(), service)
-        writer.write("}")
+        writer.writeLine("}")
```
jvmusin, line 72: This is the final line of the file. Previously it was just the closing curly bracket, now it also includes a newline character.

Again, this is the only behavioral change in this PR.

Mr3zee:
@jvmusin Thanks for the first contribution! This does indeed make the code more readable
I'll approve the changes once CI check are complete

Mr3zee (approved):

PR #78, opened 2024-05-24
Upgrade to Kotlin 2.0.0

DRSchlaubi:


e5l:
Hey @DRSchlaubi,  thanks for the PR, we'll check.

Mr3zee:
Hey, @DRSchlaubi ! Thank you for the interest in the library and your effort the make it better! Unfortunately, I will need to postpone the update to `2.0.0` as it requires complete removal of KSP plugin and migration to Kotlin Compiler Plugin. Good news is, I'm already working on it, but it will be delivered with a delay, because compiler plugins require a lot of attention.

DRSchlaubi:
Just out of curiosity, why does KSP need to be removed, all the tests passed fine?

Mr3zee:
It's not just for 2.0.0, we are removing KSP from the whole project. Where are a couple of reasons: not working gradle configuration caches, no ability to hide generated code, necessity to explicitly add KSP gradle plugin. I didn't measure, but i suspect that KSP is also slower than kotlin compiler plugin

lppedd:
Happy to see you're moving to a compiler plugin. They seem like what the community will mostly work with in the future, if docs gets taken care of. KSP is nice for delivering fast, but getting rid of an additional build time dependency is always good.  

OT, but my only doubt with compiler plugins is about processing order. If more and more plugins start popping up we might end up using many in our build, and I'm not sure if they'd interfere with each other.

Mr3zee:
@lppedd indeed the order of work for compiler plugins is what we are also looking into. Of course, not everything can be taken into account, but we will at least ensure our plugins work smoothly with very popular ones, like kotlinx.serialization

PR #86, opened 2024-05-30
Docs: Add structure to existing content

vnikolova:
**Subsystem**
Docs

**Problem Description**
Currently there's only a large Getting started topic in docs.

**Solution**
Split up the document into separate XML topics.

1. Get started - includes an introduction to the library and an installation guide
2. Plugins - explains the differences between the two plugins provided by the library
3. Core concepts - the main concepts and terminology explained with relation to the API
  3.1. Services
  3.2. RPC Clients
  3.3. RPC Servers
  3.4. Transport
4. Configuration - usage of RPCConfig
5. Features - send and receive flows, service fields, field declarations
6. Releases - release and versioning

Can you please check if this structure makes sense?


Mr3zee (changes requested):
Hi, @vnikolova ! Thank you for the update!
I just want to suggest a bit restructuring the topics, like this:

1. Get started
2. Plugins
3. Core concepts
3.1. Services
3.2. RPC Clients
3.3. RPC Servers
4. Protocols
4.1 kRPC
4.1.1 Configuration
4.1.2 Features
4.1.3 Transport 
5. Releases - release and versioning

The reason for such a structure, is that `Configuration`, `Features` (in their current way) and `Transport` are only applicable to kRPC protocol. Ans also there are plans for other protocols support, so we can add them to the `Protocols` section

Please tell me if this makes sense to you

vnikolova:
@Mr3zee thank you for clarifying. It makes perfect sense!

Mr3zee (approved):
Thank you for the PR!

PR #92, opened 2024-06-04
Update the project structure to use a new versioning

Mr3zee:
- Feature version for runtime modules
- Kotlin prefix version for compiler plugins


Mr3zee:
Tests are passing, checked one by one on the TC, GH does not keep test results when you commit new changes

e5l (approved):
LGTM. Gradle is tough :D

PR #98, opened 2024-06-13
IDEA. Vector icon

turansky:
**Problem Description**
Non-vector icon

**Solution**
Use vector icon


Mr3zee:
@turansky what is the difference for IDE here?

turansky:
From UI you can choose [SVG only](https://www.jetbrains.com/help/idea/customize-actions-menus-and-toolbars.html#change-project-icon).
Icon can be use used in different places (IDE and OS) with different scale - vector format is preferred in such case.

Mr3zee (approved):

PR #99, opened 2024-06-13
Fixed GH Actions for docs

Mr3zee:
Previous approach with reusable workflows didn't work, so I added `if` statements to `deploy` and `publish-indexes` instead:

```
if: github.event_name == 'push' && github.ref == 'refs/heads/main'
```


e5l (approved):
let's try it now

PR #104, opened 2024-06-19
KRPC-59 streamScoped has internal package

Mr3zee:
**Subsystem**
Core

**Problem Description**
`streamScoped` and `invokeOnStreamScopeCompletion` public APIs were wrongly located in `kotlinx.rpc.internal` package.

**Solution**
Move to `kotlinx.rpc` package, deprecate old methods


e5l (approved):
lgtm.

I think we need to check if we can avoid using top-level functions for that

PR #105, opened 2024-06-19
KRPC-18 Add K2 and IR code generation plugins, preserve KSP for K1

Mr3zee:
Adds support for K2 code generation. Uses K2 compiler plugin instead of KSP.
Backend Compiler Plugin now generates most of the code, FIR plugin for K2 and KSP for K1 only generate minimal declarations needed for Serialization and metadata generation on the frontend.

This is a meta-PR that accumulated many other PRs, see corresponding commits and PR descriptions.


efemoney:
Curious, is there more information on this change? Motivation, effect, etc

Mr3zee:
@efemoney basically these reasons:

- KSP Gradle plugin needs to be added explicitly by the user, that's a friction in UX
- KSP does not support Gradle configuration caches
- KSP exposes generated code, which we often don't want to do to keep the sources clean
- KSP does not allow for existing code modifications, so we need a compiler plugin anyway
- K2 frontend compiler plugins will allow for IDE integration for generated declarations in future, so we might make use of it too

efemoney:
Makes sense! Thanks

e5l (approved):
Hey! Good job! It also would be great to mention YT ticket in the header and update description

Mr3zee:
@e5l thank you! Yes, forgot to do it, well do soon

Issue #93, opened 2024-06-06
[iOS] Uncaught Kotlin exception: kotlin.IllegalStateException: Internal kotlinx.rpc error: unable to find class UserService rpc client object - on Sampled app


jerryOkafor:
I tried following the Readme guidelines and ended up with the following error:

```
Uncaught Kotlin exception: kotlin.IllegalStateException: Internal kotlinx.rpc error: unable to find class UserService rpc client object
```

To be clear, Android works as expected and to be sure I am not making any mistakes, I pulled the sample, tried it on iOS, and ended up with the same issue.

Environment:
Kotlin: 1.9.24
Xcode Version : 15.4 (15F31d)

Mr3zee:
Thanks for the report! I will investigate

afTrolle:
I'm also having the similar issue, However I have a different project setup than the sample.

I moved the RPC interface to it's own module that then the server and client modules depend on.

`kotlin.IllegalStateException: Internal kotlinx.rpc error: unable to find class run.tevla.api.ForgeServiceContract rpc client object`

![image](https://github.com/Kotlin/kotlinx-rpc/assets/1321086/d85054b3-bef7-44af-b19d-6d3ec411369a)

jerryOkafor:
This is similar to how my modules are arranged. When I noticed the issue, I pulled the sample and ran it to verify that the same issue existed.

ShayOinif:
Hello.
I am facing the same issue and would love to have a fix for it.
Thanks!

ShayOinif:
Seems like the problem is present on macOS targets as well.
Thought maybe I misconfigured my gradles, but in the sample from the code base the issue reoccurred.

ShayOinif:
Tested and found the same issue on Linux targets as well.

xxfast:
Only seems to be affecting the native targets - JVM and Android targets works as expected 👍

Mr3zee:
Fixed in #112

PR #112, opened 2024-06-24
Fix native targets compiler plugins (#93)

Mr3zee:
**Subsystem**
RPC Gradle plugin

**Problem Description**
Transitive dependencies in compiler plugin artifacts do not work for Kotlin/Native

**Solution**
We have only a few transitive dependencies, so I hardcoded them


gradle-plugin/gradle-plugin-api/src/main/kotlin/kotlinx/rpc/RPCPluginConst.kt
```diff
@@ -11,28 +11,27 @@ import org.jetbrains.kotlin.gradle.utils.loadPropertyFromResources
...
     const val GROUP_ID = "org.jetbrains.kotlinx"
 
-    const val COMPILER_PLUGIN_MODULE = "${GROUP_ID}:compiler-plugin"
     const val KSP_PLUGIN_MODULE = "${GROUP_ID}:ksp-plugin"
 
     const val KSP_PLUGIN_ID = "com.google.devtools.ksp"
 
     const val PLUGIN_ID = "org.jetbrains.kotlinx.rpc.plugin"
-    const val COMPILER_PLUGIN_ARTIFACT_ID = "kotlinx-rpc-compiler-plugin"
+    var COMPILER_PLUGIN_ARTIFACT_ID = "kotlinx-rpc-compiler-plugin"
```
e5l, line 19: Could we make a function to compute value instead?

Mr3zee: sure

PR #116, opened 2024-06-25
K2 plugin

Mr3zee:
**Subsystem**
K2 Compiler plugin

**Problem Description**
We are migrating from KSP code generation to compiler plugins. 

**Solution**
This is part 2 of 3 in the migration process, which presents the K2 frontend plugin.

K2 plugin overview:
- Detects RPC services
- Generated minimal service stub implementation class and its companion object
- Generates minimal method classes declarations
- Executes serialization plugin manually for generated classes

Also PR contains addition of an `compiler-plugin-common` module for shared declarations between backend and frontend(s).

PR contains upgrade to Kotlin 2.0 as it required to make K2 plugin work. Project language level is set to 1.7 by default, so K2 will not be used by default. For it to be used in tests, use dedicated `kotlinx.rpc.useK2Plugin=true` in root `gradle.properties`


Mr3zee:
No tests will be executed for this review for two reasons:
- No compiler plugin tests are present atm (will be added in the subsequent PRs)
- Versions earlier than 2.0 will fail due to broken KSP (will be fixed in the subsequent PR)

gradle-conventions/compiler-specific-module/src/main/kotlin/compiler-specific-module.gradle.kts
```diff
@@ -62,15 +62,17 @@ fun KotlinSourceSet.configureResources(sourceSetPath: Path, vararg versionNames:
...
-        from(versionNames.map { vsResources.resolve(it) })
-        include { it.file.parentInAllowList(versionNames) }
-    }
+    // todo duplicate (or to many resources are copied, should update the algo)
+//    tasks.withType<ProcessResources>().configureEach {
+//        duplicatesStrategy = DuplicatesStrategy.EXCLUDE
+//
+//        from(versionNames.map { vsResources.resolve(it) })
+//        include { it.file.parentInAllowList(versionNames) }
+//    }
```
Mr3zee, line 71: I will fix it in a separate PR

compiler-plugin/compiler-plugin-k2/src/main/core/kotlinx/rpc/codegen/FirRPCServiceGenerator.kt
```diff
@@ -0,0 +1,436 @@
...
+            rpcMethodClassKey != null -> {
+                when {
+                    !rpcMethodClassKey.isObject -> setOf(
+                        SpecialNames.DEFAULT_NAME_FOR_COMPANION_OBJECT,
+                        SerialEntityNames.SERIALIZER_CLASS_NAME,
+                    )
+
+                    // otherwise an object is generated instead of a class
+                    // serialization plugin has other logic for such declarations
+                    else -> setOf()
```
e5l, line 136: emptySet()

e5l (approved):
Let's get someone from the compiler team to check it out as well

Mr3zee:
@e5l do you have someone specific in mind?

compiler-plugin/compiler-plugin-k2/src/main/core/kotlinx/rpc/codegen/FirRPCServiceGenerator.kt
```diff
@@ -0,0 +1,436 @@
...
+     * If the [owner] is an RPC service - generates its service stub.
+     * Scrapes the functions from the [owner] to generate method classes.
+     */
+    private fun generateRpcServiceStubClass(owner: FirClassSymbol<*>): FirRegularClassSymbol? {
+        owner.resolvedSuperTypes.find {
+            it.classId == ClassDeclarations.rpcInterface
+        } ?: return null
+
+        val functions = mutableListOf<FirFunctionSymbol<*>>()
+        owner.declaredMemberScope(session, null).processAllFunctions {
```
demiurg906, line 281: This place is suspicious. It's not the best idea to call `declaredMemberScope` inside the generation extension, as it will return the scope, which will include explicitly declared functions **and** functions generated by plugins, which may bring your extension into inconsistent state on in the recursion

In `getCallable...` methods there is a `context.declaredScope`, which contains only source-declared declarations, which is expected to use in such places

compiler-plugin/compiler-plugin-k2/src/main/core/kotlinx/rpc/codegen/FirRPCServiceGenerator.kt
```diff
@@ -0,0 +1,436 @@
...
+ * - [generateSerializerImplClass]
+ *
+ * This is basically copying the behavior of the actual plugin but isolated only for our generated classes.
+ */
+class FirRPCServiceGenerator(
+    session: FirSession,
+    @Suppress("unused")
+    private val logger: MessageCollector,
+) : FirDeclarationGenerationExtension(session) {
+    private val serializationExtension = SerializationFirResolveExtension(session)
```
demiurg906, line 85: The trick with your own generator from serialization plugin is nice
Like it

compiler-plugin/compiler-plugin-k2/src/main/core/kotlinx/rpc/codegen/FirRPCServiceGenerator.kt
```diff
@@ -0,0 +1,436 @@
...
+            visibility = Visibilities.Public
+
+            rpcMethodClassKey.rpcMethod.valueParameterSymbols.forEach { valueParam ->
+                valueParameter(
+                    name = valueParam.name,
+                    type = valueParam.resolvedReturnType,
+                )
+            }
+        }.also {
+            it.containingClassForStaticMemberAttr = ConeClassLikeLookupTagImpl(context.owner.classId)
```
demiurg906, line 361: `createConstructor` already does it

compiler-plugin/compiler-plugin-k2/src/main/core/kotlinx/rpc/codegen/FirRPCServiceGenerator.kt
```diff
@@ -0,0 +1,436 @@
...
+            rpcMethodClassKey != null -> {
+                getCallableNamesForRpcMethodClass(classSymbol, context, rpcMethodClassKey)
+            }
+
+            classSymbol.isFromSerializationPlugin -> {
+                serializationExtension.getCallableNamesForClass(classSymbol, context)
+            }
+
+            else -> {
+                super.getCallableNamesForClass(classSymbol, context)
```
demiurg906, line 304: What's the point to call `super` implementation, if it returns an empty set either way?
Same for all other places like this

demiurg906 (commented):
Generally, the approach is fine

But I recall that we discussed that the plugin also should check some invariants with checkers, and I don't see them here

Will they be implemented later, or I confused it with some other plugin?

compiler-plugin/compiler-plugin-k2/src/main/core/kotlinx/rpc/codegen/FirRPCServiceGenerator.kt
Mr3zee: @demiurg906 Well, I need this list in getNestedClassfiersNames, because I generate a class for each function (roughly)
Is there any other API I could use for this?

Mr3zee:
@demiurg906 thanks for the review!

Yes, you're right, we talked about checkers, they will be shipped separately

compiler-plugin/compiler-plugin-k2/src/main/core/kotlinx/rpc/codegen/FirRPCServiceGenerator.kt
demiurg906: `owner.fir.declarations` will be safer
Just opt-in to `@SymbolInternals` and don't save functions themselves, only symbols

PR #123, opened 2024-07-01
Move all Gradle plugins' code to the one module

Mr3zee:
**Subsystem**
Gradle Plugin

**Problem Description**
Gradle Plugin portal can not host non-plugin dependencies, like `kotlinx-rpc-gradle-plugin-api`. And the whole 3-module structure makes little sense. 

**Solution**
Gather all gradle plugins' code into one module: no dependencies - no problems


gradle-plugin/build.gradle.kts
```diff
@@ -3,53 +3,91 @@
...
+
+dependencies {
+    compileOnly(libs.kotlin.gradle.plugin)
+}
+
+// This block is needed to show plugin tasks on --dry-run
+//  and to not run task actions on ":plugin:task --dry-run".
+//  The bug is known since June 2017 and still not fixed.
+//  The workaround used below is described here: https://github.com/gradle/gradle/issues/2517#issuecomment-437490287
+if (gradle.parent != null && gradle.parent!!.startParameter.isDryRun) {
```
e5l, line 29: you can simplify it to (?. chain == true)

e5l (approved):

PR #128, opened 2024-07-10
Add WebSocket installation to snippet

SebastianAigner:
Since otherwise, trying to run that snippet throws an Exception on startup that WebSocket is missing :)


Mr3zee:
Hi, @SebastianAigner ! This was fixed in `0.2.1` I believe, so we can close this PR.

PR #139, opened 2024-07-17
Fix log error messages

Mr3zee:
**Subsystem**
kRPC

**Problem Description**
A couple of problems with logs and messages order

**Solution**
- Introduce cancellation acknowledgement to keep messages from races and subsequent errors
- ignore cancellations in logs


Mr3zee:
All the bugs are brought to you by the "Concurrency ©", so it is really hard to test it without stress tests, which we don't have for now

e5l (approved):
LGTM! Please fix CI before merging

PR #149, opened 2024-08-04
Restructure compiler-plugin module

Mr3zee:
**Subsystem**
Compiler plugin

**Problem Description**
`compiler-plugin` has wrong structure of submodules.

**Solution**
New structure mirrors one that is dominant in `Kotlin` repository:
- backend - for IR extension plugin
- common - for common declarations
- k2 - for K2 FIR plugin
- cli - for Plugin registrar, entry point for the whole plugin


compiler-plugin/compiler-plugin-cli/src/main/core/Stub.kt
```diff
@@ -0,0 +1,11 @@
+/*
```
e5l, line 1: copyright twice

e5l (approved):

PR #152, opened 2024-08-05
Fix CSM resources load for all Kotlin versions

Mr3zee:
**Subsystem**
Compiler specific modules

**Problem Description**
Resources were loaded improperly for older Kotlin versions

**Solution**
We have CSMs (Compiler specific modules), which load different sources sets depending on the current Kotlin version. Same applies for resources, but it didn't work for them. This PR fixes it.


build.gradle.kts
```diff
@@ -15,6 +15,13 @@ plugins {
     alias(libs.plugins.binary.compatibility.validator)
 }
 
+// useful for dependencies introspection
+// run ./gradlew htmlDependencyReport
+// Report can normally be found in build/reports/project/dependencies/index.html
+allprojects {
+    plugins.apply("project-report")
```
e5l, line 22: oh that's nice

e5l (approved):

PR #155, opened 2024-08-06
Update project for K1 usage

Mr3zee:
**Subsystem**
Compiler plugins, Gradle configurations

**Problem Description**
While implementing K2, some temporary adjustments were made, disabling K1 compilation, while making K2 development process easier.

**Solution**
Enable K1 back (via KSP), but now alongside with K2


Mr3zee:
It's TC config problems, will address soon

e5l (approved):

PR #170, opened 2024-08-19
Exception Deserialization

Mr3zee:
**Subsystem**
Core

**Problem Description**
Exceptions with message and cause are not deserialized properly 

**Solution**
- Fix constructor choosing when deserializing
- Fix field counting


krpc/krpc-test/src/jvmMain/kotlin/kotlinx/rpc/test/KRPCTransportTestBase.kt
```diff
@@ -370,7 +371,6 @@ abstract class KRPCTransportTestBase {
         }
     }
 
-    @OptIn(ExperimentalCoroutinesApi::class)
     @Test
     @Suppress("detekt.TooGenericExceptionCaught")
     fun testException() {
```
e5l, line 376: name could have a bit more details

e5l (approved):

PR #181, opened 2024-08-27
Fix regression of nested declarations in RPC interfaces

Mr3zee:
**Subsystem**
Compiler plugin

**Problem Description**
We introduced regression in patch release `0.2.4` that breaks compilation in case of nested declarations in RPC interfaces

**Solution**
Switch the error to a warning

### DIFF
--- a/compiler-plugin/compiler-plugin-backend/src/main/core/kotlinx/rpc/codegen/extension/RPCDeclarationScanner.kt
+++ b/compiler-plugin/compiler-plugin-backend/src/main/core/kotlinx/rpc/codegen/extension/RPCDeclarationScanner.kt
@@ -5,6 +5,8 @@
 package kotlinx.rpc.codegen.extension
 
 import kotlinx.rpc.codegen.common.RpcNames
+import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity
+import org.jetbrains.kotlin.cli.common.messages.MessageCollector
 import org.jetbrains.kotlin.ir.declarations.IrClass
 import org.jetbrains.kotlin.ir.declarations.IrDeclaration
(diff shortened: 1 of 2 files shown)


compiler-plugin/compiler-plugin-backend/src/main/core/kotlinx/rpc/codegen/extension/RPCDeclarationScanner.kt
```diff
@@ -90,6 +92,11 @@ internal object RPCDeclarationScanner {
     }
 }
 
-private fun unsupportedDeclaration(service: IrClass, declaration: IrDeclaration): Nothing {
-    error("Unsupported declaration in RPC interface ${service.name}: ${declaration.dumpKotlinLike()}")
+private fun unsupportedDeclaration(service: IrClass, declaration: IrDeclaration, logger: MessageCollector): Nothing? {
```
e5l, line 95: I would skip the explicit return type here and drop the return null here, but not sure about map return type

e5l (approved):
lgtm

compiler-plugin/compiler-plugin-backend/src/main/core/kotlinx/rpc/codegen/extension/RPCDeclarationScanner.kt
Mr3zee: IDE only suggests `Unit?` instead, which I don't think is a good alternative

Mr3zee: Something like `private fun unsupportedDeclaration(...) = null` automatically infers to `Nothing?`, so I guess it's ok the leave it as is

PR #191, opened 2024-09-11
Added proxy repository settings

Mr3zee:
**Subsystem**
Gradle

**Problem Description**
We used direct access for our package repositories, which may result in delays and throttling (and sometimes 500 codes, I'm looking at you npm)

**Solution**
Use proxy caching repositories

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,4 +35,5 @@ detekt/reports/**
 # Ignore kover reports
 /kover
 
+local.properties
 scan-journal.log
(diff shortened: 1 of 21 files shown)


e5l (approved):
Could you tell me how it affected the build time?

Mr3zee:
@e5l it's around the same, [new](https://ge.jetbrains.com/s/mouhhl6mixlps/console-log?page=2) vs [old](https://ge.jetbrains.com/s/5tyaw2govmd7q/console-log?anchor=994&page=1)

PR #192, opened 2024-09-12
Added build cache and develocity plugin

Mr3zee:
**Subsystem**
Gradle

**Solution**
Added Develocity build cache and scans

### DIFF
--- a/gradle-conventions-settings/develocity/src/main/kotlin/conventions-develocity.settings.gradle.kts
+++ b/gradle-conventions-settings/develocity/src/main/kotlin/conventions-develocity.settings.gradle.kts
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2023-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+import java.util.*
+
+plugins {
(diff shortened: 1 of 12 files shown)


gradle-conventions-settings/develocity/src/main/kotlin/conventions-develocity.settings.gradle.kts
```diff
@@ -0,0 +1,52 @@
...
+
+        capture {
+            fileFingerprints = true
+        }
+
+        buildScanPublished {
+            scanJournal.appendText("${Date()} — $buildScanUri — $startParameter\n")
+        }
+
+        publishing.onlyIf { true }
```
CristianGM, line 34: This is the only thing I'm worried. Every external contributor will try publish scans and fail with a reject.

Mr3zee: @CristianGM I will introduce a property to disable publication attempt

e5l (approved):
Could you also add a similar one to Ktor?

Mr3zee:
@e5l check https://github.com/ktorio/ktor/pull/4311

PR #208, opened 2024-09-25
Fixed Wasm Publication

Mr3zee:
**Subsystem**
Gradle

**Problem Description**
Wasm artifacts were published without `kotlinx-rpc-` prefix

**Solution**
Add missing configs


e5l (approved):
lgtm

PR #214, opened 2024-10-08
Experimental support for: KRPC-125 Manual stream scope management

Mr3zee:
**Subsystem**
kRPC

**Problem Description**
`streamScoped` function is not always usable, due to its strict lifetime limits

**Solution**
Provide `StreamScope` resource that is manageable via manual `close` calls.

### DIFF
--- a/krpc/krpc-core/src/commonMain/kotlin/kotlinx/rpc/krpc/StreamScope.kt
+++ b/krpc/krpc-core/src/commonMain/kotlin/kotlinx/rpc/krpc/StreamScope.kt
@@ -15,6 +15,7 @@ import kotlin.contracts.InvocationKind
 import kotlin.contracts.contract
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.coroutineContext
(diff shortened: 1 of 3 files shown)


krpc/krpc-core/src/commonMain/kotlin/kotlinx/rpc/krpc/StreamScope.kt
```diff
@@ -165,22 +179,53 @@ public suspend fun <T> streamScoped(block: suspend CoroutineScope.() -> T): T {
...
+        }
+    }
+}
 
-    if (context[StreamScope.Key] != null) {
+private fun CoroutineContext.checkContextForStreamScope() {
+    if (this[StreamScope.Element] != null) {
         error(
             "One of the following caused a failure: \n" +
                     "- nested 'streamScoped' calls are not allowed.\n" +
```
e5l, line 199: should we also mention withStreamScope?

e5l (approved):
Well done, lgtm!

PR #217, opened 2024-10-30
Support Kotlin 2.0.21 and 2.0.20

Mr3zee:
**Subsystem**
Gradle

**Problem Description**
2.0.20 and 2.0.21 Kotlin versions are out, and we don't support them yet.

**Solution**
Now with #216 merged, we can support them easily


e5l (approved):
please check why other builds are not triggered

Mr3zee:
@e5l I already triggered them (all build in the projects and manual). After pushing a new commit GH deleted builds from the checks list. But you can see on TC, for example, that they are there and good

PicoPocoo:
Looking forward to the next release, where 2.0.21 is supported. Thanks for your great work!

PR #227, opened 2024-11-05
Update Service Descriptors Generation

Mr3zee:
**Subsystem**
Core, Compiler Plugin

**Problem Description**
Current service descriptor API is not complete and overcomplicated. It lacks some info about the service, as well as some functionality

**Solution**
Introduce Descriptor API:
- Everything is contained within an `RpcServiceDescriptor` with a single entrypoint - `serviceDescriptorOf`
- Added API to call a field or a method
- Exhaustive type information on methods and fields
- Simplified `RpcClient` API


core/src/commonMain/kotlin/kotlinx/rpc/RpcClient.kt
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * This method is used by generated clients to perform a call to the server.
+     *
+     * @param T type of the result
+     * @param serviceScope service's coroutine scope
+     * @param call an object that contains all required information about the called method,
+     * that is needed to route it properly to the server.
+     * @return actual result of the call, for example, data from the server
+     */
+    public fun <T> callSync(serviceScope: CoroutineScope, call: RpcCall): Deferred<T>
```
e5l, line 36: I would name it `callBlocking`

tests/compiler-plugin-tests/src/main/kotlin/kotlinx/rpc/codegen/test/TestRpcClient.kt
```diff
@@ -5,34 +5,40 @@
...
-        return MutableSharedFlow<String>(1).also {
-            GlobalScope.launch { it.emit("registerSharedFlowField_42") }
-        } as SharedFlow<T>
-    }
+    override fun <T> callSync(serviceScope: CoroutineScope, call: RpcCall): Deferred<T> {
+        val callable = call.descriptor.getCallable(call.callableName)
+            ?: error("No callable found for ${call.callableName}")
+
+        val value = when (callable.name) {
+            "plainFlow" -> flow { emit("registerPlainFlowField_42") }
```
e5l, line 30: please consider introducing enum for all values

e5l (approved):

tests/compiler-plugin-tests/src/main/kotlin/kotlinx/rpc/codegen/test/TestRpcClient.kt
Mr3zee: It's a test class for code generation, I don't think it is needed there, enum will need to be updated the same way the string consts here

core/src/commonMain/kotlin/kotlinx/rpc/RpcClient.kt
Mr3zee: Maybe `callInScope`? It is actually not sync neither blocking

whyoleg: what about `callAsync`? it returns `Deferred` in the same way as `async` from `kotlinx.coroutines` returns `Deferred`
Also: https://kotlinlang.org/docs/composing-suspending-functions.html#async-style-functions

core/src/commonMain/kotlin/kotlinx/rpc/RpcClient.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Provides child [CoroutineContext] for a new [RemoteService] service stub.
+     *
+     * This function shouldn't be called directly.
+     *
+     * @param serviceId id of the new service. Used for service cancellation messages.
+     */
+    public fun provideStubContext(serviceId: Long): CoroutineContext
```
whyoleg, line 45: curious: why it's called like this and not something like `createRemoteServiceContext`?
AFAIU it will be used for `RemoteService` client implementations, so seems legit
(probably I just don't like `stub` in API naming)

core/src/commonMain/kotlin/kotlinx/rpc/descriptor/RpcServiceDescriptor.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    internalError(
+        "Located descriptor object is not of a desired type ${RpcServiceDescriptor::class}, " +
+                "instead found $maybeDescriptor of the class " +
+                (maybeDescriptor::class.qualifiedClassNameOrNull ?: maybeDescriptor::class)
+    )
+}
+
+@ExperimentalRPCApi
+public interface RpcServiceDescriptor<T : RemoteService> {
```
whyoleg, line 42: by reading this declaration I do have a strange feeling, that it should be called `RemoteServiceDescriptor`, or `RemoteService` should be called `RpcService`.

core/src/commonMain/kotlin/kotlinx/rpc/descriptor/RpcServiceDescriptor.kt
```diff
@@ -0,0 +1,75 @@
...
+public fun <T : RemoteService> serviceDescriptorOf(kType: KType): RpcServiceDescriptor<T> {
+    return serviceDescriptorOf(kType.kClass())
+}
+
+@ExperimentalRPCApi
+public fun <T : RemoteService> serviceDescriptorOf(kClass: KClass<T>): RpcServiceDescriptor<T> {
+    val maybeDescriptor = internalServiceDescriptorOf(kClass)
+        ?: internalError("Unable to find a service descriptor of the $kClass")
+
+    if (RpcServiceDescriptor::class.isInstance(maybeDescriptor)) {
```
whyoleg, line 29: curious: why `maybeDescriptor is RpcServiceDescriptor` doesn't work here?

core/src/commonMain/kotlin/kotlinx/rpc/descriptor/RpcServiceDescriptor.kt
```diff
@@ -0,0 +1,75 @@
...
+                "instead found $maybeDescriptor of the class " +
+                (maybeDescriptor::class.qualifiedClassNameOrNull ?: maybeDescriptor::class)
+    )
+}
+
+@ExperimentalRPCApi
+public interface RpcServiceDescriptor<T : RemoteService> {
+    public val fqName: String
+
+    public fun getFields(service: T): List<RpcDeferredField<*>>
```
whyoleg, line 45: it refers to a class in `internal` package with `internal` annotation - is it correct or should it also moved here?

Also, may be there is some information about the design of `RpcDeferredField` on why it's needed and why it's done like this? Couldn't find it :(

whyoleg (commented):
just some small questions to better understand the idea as a whole :)

core/src/commonMain/kotlin/kotlinx/rpc/RpcClient.kt
Mr3zee: Well, stub is the official name for generated service implementations, thus this naming explicitly says what it will do. It's the why

In terms of can we do better, probably yes, but it's not the scope of this PR, it was moved as is from the old `RPCClient`

core/src/commonMain/kotlin/kotlinx/rpc/descriptor/RpcServiceDescriptor.kt
Mr3zee: Yeah, we have a problem with naming here. 
We hopefully will address the necessity of the `RemoteService` on our design session, and it may be removed at all, so `@Rpc` annotation and `RpcServiceDescriptor` will be in synergy

core/src/commonMain/kotlin/kotlinx/rpc/descriptor/RpcServiceDescriptor.kt
Mr3zee: No idea 😅 
I just did it automatically like this, is there a good reason why `in` might be better?

core/src/commonMain/kotlin/kotlinx/rpc/descriptor/RpcServiceDescriptor.kt
Mr3zee: Oh, that's a good catch. It probably should also be internal.

Concerning the design, it was done like to just work. We will address the necessity of fields on out design session, I will provide a doc with current state and comments beforehand

core/src/commonMain/kotlin/kotlinx/rpc/descriptor/RpcServiceDescriptor.kt
whyoleg: using `is` is more kotlin-ish
`KClass::isInstance` is mostly needed when you have `KClass` object passing around, but here it just doesn't make sense as we check against known type :)

PR #233, opened 2024-11-18
Support Kotlin 2.1.0-RC2

Mr3zee:
**Subsystem**
Compiler Plugin

**Problem Description**
Kotlin `2.1.0-RC` is out with breaking changes in Compiler API

**Solution**
Update and fix changes

### DIFF
--- a/gradle-conventions-settings/src/main/kotlin/conventions-repositories.settings.gradle.kts
+++ b/gradle-conventions-settings/src/main/kotlin/conventions-repositories.settings.gradle.kts
@@ -5,6 +5,21 @@
 @file:Suppress("DuplicatedCode")
 
 pluginManagement {
+    fun findGlobalRootDirPath(): java.nio.file.Path {
+        var path = file(".").toPath().toAbsolutePath()
+
+        // we assume that the 'versions-root' folder can only be present in the root folder
+        while (
+            java.nio.file.Files.newDirectoryStream(path).use { it.toList() }.none {
+                java.nio.file.Files.isDirectory(it) && it.fileName.toString() == "versions-root"
+            }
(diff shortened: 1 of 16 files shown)


versions-root/libs.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
 # core library version
-kotlinx-rpc = "0.4.0"
+kotlinx-rpc = "0.5.0"
```
e5l, line 3: This would be nice to have in the separate commit.
Should we use `0.5.0-SNAPSHOT` instead?

gradle-conventions-settings/src/main/kotlin/conventions-repositories.settings.gradle.kts
```diff
@@ -80,10 +95,27 @@ pluginManagement {
             mavenCentral()
             gradlePluginPortal()
         }
+
+        maven("${findGlobalRootDirPath()}/lib-kotlin/")
     }
 }
 
 gradle.rootProject {
+    fun findGlobalRootDirPath(): java.nio.file.Path {
```
e5l, line 104: could be lazy val

e5l (approved):

gradle-conventions-settings/src/main/kotlin/conventions-repositories.settings.gradle.kts
Mr3zee: Would be the same, I instead used it once for invocation

versions-root/libs.versions.toml
Mr3zee: Made as a separate commit

PR #234, opened 2024-11-19
Update the project structure to work with kotlin-master

Mr3zee:
**Subsystem**
Gradle

**Problem Description**
Kotlin master build require local artifacts, which don't work for current project structure, as it can't resolve KGP this way.

**Solution**
Change the structure to work with local kotlin artifacts (move KGP dependent files from settings to plain conventions)


gradle-conventions-settings/src/main/kotlin/util/KotlinVersionActions.kt
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+package util
+
+import org.gradle.api.plugins.ExtensionAware
+import org.gradle.kotlin.dsl.extra
+import org.gradle.kotlin.dsl.provideDelegate
+
+
+data class ActionApplied(val applied: Boolean)
```
e5l, line 12: I would use a plain boolean or introduce an enum if readability is needed.

PR #239, opened 2024-11-25
Updated descriptor to use `RpcType` instead of `KType` directly

Mr3zee:
**Subsystem**
Core

**Problem Description**
KType maybe changes in the future to KClass or custom metadata

**Solution**
Introduce intermediate `RpcType`

### DIFF
--- a/compiler-plugin/compiler-plugin-backend/src/main/core/kotlinx/rpc/codegen/extension/RpcIrContext.kt
+++ b/compiler-plugin/compiler-plugin-backend/src/main/core/kotlinx/rpc/codegen/extension/RpcIrContext.kt
@@ -105,6 +105,10 @@ internal class RpcIrContext(
         getRpcIrClassSymbol("RpcServiceDescriptor", "descriptor")
     }
 
+    val rpcType by lazy {
+        getRpcIrClassSymbol("RpcType", "descriptor")
+    }
+
     val rpcCallable by lazy {
         getRpcIrClassSymbol("RpcCallable", "descriptor")
     }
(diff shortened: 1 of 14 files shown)


compiler-plugin/compiler-plugin-backend/src/main/core/kotlinx/rpc/codegen/extension/RpcIrContext.kt
```diff
@@ -105,6 +105,10 @@ internal class RpcIrContext(
         getRpcIrClassSymbol("RpcServiceDescriptor", "descriptor")
     }
 
+    val rpcType by lazy {
```
e5l, line 108: Should we extract type constants to the enum?

gradle-conventions/latest-only/src/main/kotlin/conventions-gradle-doctor.gradle.kts
```diff
@@ -9,6 +9,6 @@ plugins {
 doctor {
     enableTestCaching = false
     warnWhenNotUsingParallelGC = true
-    disallowMultipleDaemons = true
+    disallowMultipleDaemons = false
```
e5l, line 12: memory💦 ?

e5l (approved):

e5l:
Nice, please check CI before merging

compiler-plugin/compiler-plugin-backend/src/main/core/kotlinx/rpc/codegen/extension/RpcIrContext.kt
Mr3zee: Maybe, but definitely later. It works now, and I don't have problems maintaining it.

PR #243, opened 2024-12-02
Strict mode

Mr3zee:
**Subsystem**
Core, Compiler Plugin

**Problem Description**
We are about to deprecate some features in `@Rpc` services, namely:
- StateFlow support
- SharedFlow support
- Fields
- Nested Flows
- `suspend` functions with server-side streaming 

**Solution**
We want to provide provide migration tools for users, as some of the declarations are harder to replace. That is why strict mode is introduced. For each deprecation there will be three levels of warnings: `none`, `warning` and `error`. By default, all of them are now warning expect for `suspend functions with server-side streaming`, as we are not yet ready to provide an alternative.

Deprecation level will shift to error in the following releases.
Then features will be considered obsolete, strict mode will become the default without the ability to switch if off.


compiler-plugin/compiler-plugin-k2/src/main/core/kotlinx/rpc/codegen/StrictMode.kt
```diff
@@ -0,0 +1,142 @@
...
+            return true
+        }
+    }
+
+    return false
+}
+
+private fun String.toStrictMode(): StrictMode? {
+    return when (lowercase()) {
+        "none" -> StrictMode.NONE
```
e5l, line 137: could be an extension in the enum companion

e5l (approved):

versions-root/libs.versions.toml
```diff
@@ -20,6 +20,7 @@ junit5 = "5.11.3"
 intellij = "233.13135.128"
 gradle-doctor = "0.10.0"
 kotlinx-browser = "0.2"
+shadow-jar = "9.0.0-beta2"
```
e5l, line 23: Could you tell me if this change related to the strict mode?

gradle-plugin/src/main/kotlin/kotlinx/rpc/compilerPlugins.kt
```diff
@@ -22,4 +24,26 @@ class CompilerPluginBackend : KotlinCompilerPluginSupportPlugin by compilerPlugi
...
+            ?: RpcExtension(it.target.project.objects)
+
+        val strict = extension.strict
+
+        it.target.project.provider {
+            listOf(
+                SubpluginOption("strict-stateFlow", strict.stateFlow.get().toCompilerArg()),
+                SubpluginOption("strict-sharedFlow", strict.sharedFlow.get().toCompilerArg()),
+                SubpluginOption("strict-nested-flow", strict.nestedFlow.get().toCompilerArg()),
+//                SubpluginOption("strict-stream-scope", strict.streamScopedFunctions.get().toCompilerArg()),
```
e5l, line 39: Please drop comments before merging or provide an issue for it

gradle-plugin/src/main/kotlin/kotlinx/rpc/Extensions.kt
```diff
@@ -0,0 +1,94 @@
...
+    val nestedFlow: Property<RpcStrictMode> = objects.strictModeProperty()
+
+    /**
+     * WIP
+     * Will be enabled later, when an alternative is ready.
+     */
+    private val streamScopedFunctions: Property<RpcStrictMode> = objects.strictModeProperty(RpcStrictMode.NONE)
+
+    /**
+     * WIP
```
e5l, line 63: please log an issue for WIP and link it in the comment

compiler-plugin/compiler-plugin-k2/src/main/core/kotlinx/rpc/codegen/checkers/FirRpcStrictModeClassChecker.kt
```diff
@@ -0,0 +1,202 @@
...
+            .serializableProperties
+            .mapNotNull { property ->
+                val resolvedTypeRef = property.propertySymbol.resolvedReturnTypeRef
+                val result = if (resolvedTypeRef.toClassLikeSymbol(context.session) != null) {
+                    resolvedTypeRef
+                } else {
+                    extracted[property.propertySymbol.resolvedReturnType]
+                }
+                if (result == null) {
+                    print(1)
```
e5l, line 188: Please consider introducing logger for debug purposes and drop printlns

compiler-plugin/compiler-plugin-k2/src/main/core/kotlinx/rpc/codegen/StrictMode.kt
```diff
@@ -0,0 +1,145 @@
...
+        NESTED_FLOW to StrictModeConfigurationKeys.NESTED_FLOW,
+        STREAM_SCOPED_FUNCTIONS to StrictModeConfigurationKeys.STREAM_SCOPED_FUNCTIONS,
+        SUSPENDING_SERVER_STREAMING to StrictModeConfigurationKeys.SUSPENDING_SERVER_STREAMING,
+        NOT_TOP_LEVEL_SERVER_FLOW to StrictModeConfigurationKeys.NOT_TOP_LEVEL_SERVER_FLOW,
+        FIELDS to StrictModeConfigurationKeys.FIELDS,
+    )
+}
+
+fun AbstractCliOption.processAsStrictModeOption(value: String, configuration: CompilerConfiguration): Boolean {
+    StrictModeCliOptions.configurationMapper[this]?.let { key ->
```
e5l, line 137:
```kotlin
val key = StrictModeCliOptions.configurationMapper[this] ?: return false
val mode = StrictMode.fromCli(value) ?: return false

configuration.put(key, mode)
return true
```
would be way easier to read

e5l (approved):
LGTM, please check the comments before merging

compiler-plugin/compiler-plugin-k2/build.gradle.kts
```diff
@@ -10,6 +11,49 @@ import util.whenForIde
...
+- No, because the same error now will occur in the user projects,
+  but it is com.jetbrains.intellij.PsiElement that is not found.
+
+- Ok, we can use kotlin-compiler-embeddable.jar to compile tests.
+- Then we ran into java.lang.VerifyError: Bad type on operand stack, which I have no idea how to fix.
+
+This solution replaces org.jetbrains.kotlin.com.jetbrains.intellij.PsiElement usages in plugin
+with com.jetbrains.intellij.PsiElement only for the tests, fixing both use cases.
+It is basically a reverse engineering of what Kotlin does for the embedded jar.
+ */
```
Mr3zee, line 41: Reason for the shadow jar

versions-root/libs.versions.toml
Mr3zee: Yes, it's related to testing it. See here https://github.com/Kotlin/kotlinx-rpc/pull/243/files#r1871057882

gradle-plugin/src/main/kotlin/kotlinx/rpc/compilerPlugins.kt
Mr3zee: Will add the link to the ticket

compiler-plugin/compiler-plugin-k2/src/main/core/kotlinx/rpc/codegen/checkers/FirRpcStrictModeClassChecker.kt
Mr3zee: oops

PR #254, opened 2025-01-13
Added missing aliases for migration

Mr3zee:
**Subsystem**
Core

**Problem Description**
When removing APIs, forgot to make deprecated typealiases


e5l (approved):
lgtm

PR #281, opened 2025-02-11
Simplified grpc gradle config

Mr3zee:
**Subsystem**
gRPC Gradle

**Problem Description**
gRPC Gradle config is unnecessary complicated

**Solution**
Simplify it using our Gradle plugin

### DIFF
--- a/gradle-plugin/src/main/kotlin/kotlinx/rpc/GrpcExtension.kt
+++ b/gradle-plugin/src/main/kotlin/kotlinx/rpc/GrpcExtension.kt
@@ -1,0 +1,262 @@
+/*
+ * Copyright 2023-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.rpc
+
+import com.google.protobuf.gradle.ExecutableLocator
+import com.google.protobuf.gradle.GenerateProtoTask
+import com.google.protobuf.gradle.ProtobufExtension
+import org.gradle.api.Action
+import org.gradle.api.GradleException
+import org.gradle.api.Project
+import org.gradle.api.model.ObjectFactory
+import org.gradle.api.provider.Property
+import org.gradle.api.specs.Spec
+import org.gradle.api.tasks.TaskCollection
+import org.gradle.kotlin.dsl.findByType
+import org.gradle.kotlin.dsl.property
+import org.gradle.kotlin.dsl.the
+import javax.inject.Inject
+
+public open class GrpcExtension @Inject constructor(objects: ObjectFactory, private val project: Project) {
+    /**
+     * Determines whether the gRPC support is enabled in the project.
+     *
+     * Allows for additional configuration checks.
+     */
+    public val enabled: Property<Boolean> = objects.property<Boolean>().convention(false)
+
+    /**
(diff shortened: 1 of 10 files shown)


gradle-plugin/src/main/kotlin/kotlinx/rpc/GrpcExtension.kt
```diff
@@ -0,0 +1,143 @@
...
+import org.gradle.kotlin.dsl.property
+import javax.inject.Inject
+
+open class GrpcExtension @Inject constructor(objects: ObjectFactory) {
+    /**
+     * Determines whether the gRPC support is enabled in the project.
+     *
+     * Allows for additional configuration checks.
+     */
+    val enabled: Property<Boolean> = objects.property<Boolean>().convention(false)
```
e5l, line 22: Could you tell me why do we need a separate enabled flag? Should users not install the plugin instead?

gradle-plugin/src/main/kotlin/kotlinx/rpc/GrpcExtension.kt
```diff
@@ -0,0 +1,143 @@
...
+     *
+     * ```kotlin
+     * protobuf {
+     *     plugins {
+     *         named(locatorName)
+     *     }
+     * }
+     * ```
+     */
+    val locatorName = "kotlinx-rpc"
```
e5l, line 35: Should it be Property? please check other fields

e5l (approved):
please check minor comments

protobuf-plugin/build.gradle.kts
```diff
@@ -67,37 +67,22 @@ tasks.jar {
...
+rpc {
+    grpc {
+        enabled = true
     }
+}
 
+protobuf {
     plugins {
-        create("kotlinx-rpc") {
+        maybeCreate(rpc.grpc.locatorName).apply {
```
e5l, line 78: should we put this configuration inside the plugin?

gradle-plugin/src/main/kotlin/kotlinx/rpc/GrpcExtension.kt
Mr3zee: It should not be changed, but can be accessed to manually change these configs by name

e5l: should we make accessors instead of exposing string constants then?

gradle-plugin/src/main/kotlin/kotlinx/rpc/GrpcExtension.kt
Mr3zee: It is just a flag, that allows users to verify that their configuration is proper. It is not needed per se, but can provide additional warning if something is not configured right

protobuf-plugin/build.gradle.kts
Mr3zee: It's specific only for this build, it is not needed for users

gradle-plugin/src/main/kotlin/kotlinx/rpc/GrpcExtension.kt
Mr3zee: I'll see if it is possible
But anyway, users may want to access them manually via protobuf extension

PR #287, opened 2025-02-14
Fix build config for for-ide builds

Mr3zee:
**Subsystem**
Gradle

**Problem Description**
`for-ide` artifacts were built against wrong version of the compiler, which led to incompatibilities

**Solution**
Fix configs to build against proper compiler version, respecting `kotlin-compiler` param in `libs.versions.toml`

### DIFF
--- a/compiler-plugin/compiler-plugin-backend/build.gradle.kts
+++ b/compiler-plugin/compiler-plugin-backend/build.gradle.kts
@@ -1,8 +1,10 @@
 /*
- * Copyright 2023-2024 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ * Copyright 2023-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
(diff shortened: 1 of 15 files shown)


compiler-plugin/compiler-plugin-backend/build.gradle.kts
```diff
@@ -18,6 +20,11 @@ kotlin {
 }
 
 dependencies {
-    compileOnly(libs.kotlin.compiler.embeddable)
+    whenForIde {
+        compileOnly(libs.kotlin.compiler)
+    } otherwise {
+        compileOnly(libs.kotlin.compiler.embeddable)
```
e5l, line 26: This dependency has been deprecated in the kgp. Could you tell me if we can be more specific here?

e5l (approved):

compiler-plugin/compiler-plugin-backend/build.gradle.kts
Mr3zee: `embeddable` is the proper dependency for compiler plugins, as embeddable runs during compilation process. Now, raw compiler runs in IDE, hence the difference

PR #305, opened 2025-03-17
KRPC-167 Internal symbols are poorly exposed

Mr3zee:
**Subsystem**
All

**Problem Description**
Some internal APIs have unfortunate namings, that can pop up in IDE auto completion

**Solution**
Rename the most annoying internal APIs


e5l (commented):
Should we mark them with synthetic? to drop from the completion?

Mr3zee:
@e5l You mean JvmSynthetic ?

e5l:
yep, I checked and it will not work. It would be nice to hide it somehow

Mr3zee:
@e5l I'll make a ticket for Kotlin team, we'll see what they can say

Meanwhile, we can proceed with this PR as a quick fix

e5l:
thanks, sure

e5l (approved):

Issue #315, opened 2025-04-09
ClassCastException due to Kotlin optimization


y9maly:
# Minimal reproducible example

Shared code:
```Kotlin
import kotlinx.rpc.RemoteService
import kotlinx.rpc.annotations.Rpc

@Rpc
interface MyService : RemoteService {
    suspend fun function()
}
```

Server code:
```Kotlin
import io.ktor.server.application.*
import io.ktor.server.engine.embeddedServer
import io.ktor.server.netty.Netty
import io.ktor.server.routing.routing
import kotlinx.coroutines.delay
import kotlinx.rpc.krpc.ktor.server.Krpc
import kotlinx.rpc.krpc.ktor.server.rpc
import kotlinx.rpc.krpc.serialization.json.json
import kotlin.coroutines.CoroutineContext

suspend fun doWork(): String {
    delay(100)
    return "qwerty"
}

class MyServiceImpl(override val coroutineContext: CoroutineContext) : MyService {
    override suspend fun function() {
        doWork()
    }
}

fun main() {
    embeddedServer(
        Netty,
        port = 8080,
        host = "0.0.0.0",
        module = Application::module
    ).start(wait = true)
}

fun Application.module() {
    install(Krpc) {
        serialization {
            json()
        }
    }

    routing {
        rpc("/") {
            registerService<MyService> { ctx -> MyServiceImpl(ctx) }
        }
    }
}
```

Client code:
```Kotlin
import io.ktor.client.HttpClient
import kotlinx.rpc.krpc.ktor.client.installKrpc
import kotlinx.rpc.krpc.ktor.client.rpc
import kotlinx.rpc.krpc.serialization.json.json
import kotlinx.rpc.withService

suspend fun main() {
    val client = HttpClient {
        installKrpc {
            serialization {
                json()
            }
        }
    }

    val rpc = client.rpc {
        url {
            host = "localhost"
            port = 8080
        }
    }

    val myService = rpc.withService<MyService>()

    myService.function()
}
```

# Actual result on client:

```
Exception in thread "main" java.lang.ClassCastException: class java.lang.String cannot be cast to class kotlin.Unit (java.lang.String is in module java.base of loader 'bootstrap'; kotlin.Unit is in unnamed module of loader 'app')
	at kotlinx.serialization.internal.UnitSerializer.serialize(Primitives.kt:68)
	...
```

# Workaround

```Kotlin
fun doNothing() = Unit

[...]

PR #317, opened 2025-04-10
Fix KRPC-173 (#315)

Mr3zee:
**Subsystem**
kRPC server

**Problem Description**
#315

**Solution**
Provision values for return type match


e5l (approved):

Mr3zee:
Kotlin issue: https://youtrack.jetbrains.com/issue/KT-76672/Compiler-optimization-produces-unexpected-results

PR #328, opened 2025-04-24
KRPC-143 Repeated types in gRPC

Mr3zee:
**Subsystem**
gRPC

**Solution**
Support for `repeated` types in the Protobuf plugin


e5l (approved):
nice, lgtm

PR #336, opened 2025-05-12
Remove fields support

Mr3zee:
**Strict mode Migration**
A series of changes to simplify kRPC protocol
 - [x] Remove fields support
 - [ ] Remove stream scopes
 - [ ] Simplify lifetime management
 - [ ] Cold RPC clients creation
 - [ ] Deprecated declarations cleanup
 - [ ] Documentation update for previous changes

**Subsystem**
kRPC, Core, Compiler Plugin

**Solution**
Remove fields support, as it was previously already deprecated.


e5l (approved):
Removing code feels good :D

PR #348, opened 2025-06-06
Cold Ktor client

Mr3zee:
**Strict mode Migration**
A series of changes to simplify kRPC protocol
 - [x] Remove fields support
 - [x] Remove stream scopes
 - [x] Simplify lifetime management
 - [x] Cold RPC clients creation
 - [ ] Deprecated declarations cleanup
 - [ ] Documentation update for previous changes

**Subsystem**
Krpc Ktor

**Problem Description**
Ktor Krpc client requires invokavion of `.rpc` method which is suspending. In some cases that might be bad UX.

**Solution**
Allow for a cold client creation

### DIFF
--- a/krpc/krpc-client/src/commonMain/kotlin/kotlinx/rpc/krpc/client/KrpcClient.kt
+++ b/krpc/krpc-client/src/commonMain/kotlin/kotlinx/rpc/krpc/client/KrpcClient.kt
@@ -34,6 +34,7 @@ import kotlinx.serialization.modules.SerializersModule
 import kotlin.collections.first
 import kotlin.concurrent.Volatile
 import kotlin.coroutines.cancellation.CancellationException
+import kotlin.properties.Delegates
 
 @Deprecated("Use KrpcClient instead", ReplaceWith("KrpcClient"), level = DeprecationLevel.ERROR)
 public typealias KRPCClient = KrpcClient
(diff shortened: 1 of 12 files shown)


krpc/krpc-client/src/commonMain/kotlin/kotlinx/rpc/krpc/client/KrpcClient.kt
```diff
@@ -56,15 +57,58 @@ public typealias KRPCClient = KrpcClient
...
+        checkTransportReadiness()
+
+        val context = SupervisorJob(transport.coroutineContext.job)
+
+        context.job.invokeOnCompletion(onCancelling = true) {
+            clientCancelled = true
+
+            sendCancellation(CancellationType.ENDPOINT, null, null, closeTransportAfterSending = true)
+
+            requestChannels.values.forEach { it.close(CancellationException("Client cancelled")) }
```
e5l: please avoid such operations in the invokeOnCompletion - consider launching a separate task on the global dispatcher

e5l (changes requested):

krpc/krpc-client/src/commonMain/kotlin/kotlinx/rpc/krpc/client/KrpcClient.kt
Mr3zee: Why though?

Mr3zee: And what kind of operations are we talking here? `sendCancellation` is basically what you said, closing channels should be pretty straight forward. The one thing that might be of consideration is the potential number of this channels, but I do expect it to be ~10^4-10^5 at max, which doesn't seem unreasonable

e5l (approved):
looks good

PR #354, opened 2025-06-13
Serialization decoupling

Mr3zee:
**Subsystem**
Compiler Plugin, kRPC

**Problem Description**
`kotlinx.serialization` was nailed down to our compiler plugin and core, which not always a proper coupling (like for gRPC it is unnecessary).

**Solution**
Move serialization responsibility to protocol side, switch generated objects to an array of parameters.

Note: Very are still coupled in `core` by `RpcTypeKrpc` class. Compiler plugin extensions are needed to resolve this, https://youtrack.jetbrains.com/issue/KRPC-178/General-Approach-for-custom-Service-Descriptors


utils/src/jvmMain/kotlin/kotlinx/rpc/internal/utils/map/ConcurrentHashMap.jvm.kt
```diff
@@ -20,8 +20,8 @@ private class ConcurrentHashMapJvm<K : Any, V: Any>(initialSize: Int) : RpcInter
         return map.put(key, value)
     }
 
-    override fun merge(key: K, value: V, remappingFunction: (V, V) -> V): V? {
-        return map.merge(key, value) { old, new -> remappingFunction(old, new) }
+    override fun merge(key: K, value: V, remappingFunction: (V, V) -> V): V {
+        return map.merge(key, value) { old, new -> remappingFunction(old, new) }!!
```
e5l, line 24: Please add meaningful error message

Mr3zee: There can't be null, as this map doesn't hold nulls, NPE is unreachable

PR #356, opened 2025-06-16
Platforms Table and docs

Mr3zee:
**Subsystem**
Docs

**Problem Description**
Platform status is not obvious for features in general (kRPC, gRPC) and for modules in particular.

**Solution**
Add documentations page with autogenerated platforms table


Mr3zee:
And we have working `Windows` target!

e5l (approved):

PR #360, opened 2025-06-24
Simplify Gradle Configs

Mr3zee:
**Subsystem**
Gradle

**Problem Description**
A lot of configs were left from the era when we need to compile the whole project with different Kotlin versions. Since then we introduced `compiler-plugin` separate compilation and resolution, which doesn't require the whole project to change kotlin version.

**Solution**
Simplify and prettify configs (and hopefully don't break the project)

### DIFF
--- a/.puppeteerrc.cjs
+++ b/.puppeteerrc.cjs
@@ -1,6 +1,6 @@
 const {join} = require('path');
 
-const isCI = process.env.TEAMCITY_VERSION
+const isCI = process.env.TEAMCITY_VERSION || process.env.GITHUB_ACTIONS;
 
 if (isCI) {
     /**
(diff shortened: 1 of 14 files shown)


build.gradle.kts
```diff
@@ -3,87 +3,16 @@
...
-            )
-
-            footerMessage = "© ${Year.now()} JetBrains s.r.o and contributors. Apache License 2.0"
-            homepageLink = "https://kotlin.github.io/kotlinx-rpc/get-started.html"
-
-            // replace with homepage.svg once the mentioned issue is resolved
-            templatesDir.set(templatesDirectory)
-        }
-
-        // enable versioning for stable
```
e5l, line 55: we can drop comments for now

PR #361, opened 2025-06-25
Fix Symlinks and Develocity

Mr3zee:
**Subsystem**
Gradle

**Problem Description**
- Symlinks used for `gradle.properties` don't work of Windows: https://krpc.teamcity.com/buildConfiguration/Build_kRPC_Windows/24042
- Develocity plugin didn't consider terms of use agreement

**Solution**
Fix both!

### DIFF
--- a/gradle-conventions-settings/develocity/src/main/kotlin/conventions-develocity.settings.gradle.kts
+++ b/gradle-conventions-settings/develocity/src/main/kotlin/conventions-develocity.settings.gradle.kts
@@ -16,6 +16,9 @@ develocity {
     server.set(DEVELOCITY_SERVER)
 
     buildScan {
+        termsOfUseUrl = "https://gradle.com/terms-of-service"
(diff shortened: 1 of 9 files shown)


gradle-conventions-settings/develocity/src/main/kotlin/conventions-develocity.settings.gradle.kts
```diff
@@ -35,7 +38,7 @@ develocity {
             .getOrElse("false")
             .toBooleanStrict()
 
-        publishing.onlyIf { !skipBuildScans }
+        publishing.onlyIf { termsOfUseAgree.get() == "yes" && !skipBuildScans }
```
e5l, line 41: bool?

Mr3zee: It's a built-in prop that requires yes or no

PR #370, opened 2025-06-30
Ci updates

Mr3zee:
**Subsystem**
Gradle, GA

**Problem Description**
Docs publication has some drawbacks to it:
- Long
- Hard to test

**Solution**
- Separate dokka generation into TC configs
- Add 'assemble all' step for all docs

Other updates:
- Foojay version 1.0.0
- Allow `exclude.jvm` options
- Commit changelog from docs
- Automated changelog verification
- Fix updateSitemap.sh
- Update renovate to ignore develocity

### DIFF
--- a/gradle-conventions/src/main/kotlin/conventions-root.gradle.kts
+++ b/gradle-conventions/src/main/kotlin/conventions-root.gradle.kts
@@ -2,7 +2,6 @@
  * Copyright 2023-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
-import util.other.asDokkaVersion
 import util.other.libs
 import util.tasks.configureNpm
 import util.tasks.registerChangelogTask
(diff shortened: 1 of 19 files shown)


gradle-conventions/src/main/kotlin/conventions-root.gradle.kts
```diff
@@ -63,6 +62,13 @@ dokka {
         outputDirectory = dokkaVersionsDirectory
     }
 
+    tasks.dokkaGenerate {
+        doFirst {
```
e5l, line 66: please log an issue for dokka why you need this

e5l (approved):

gradle-conventions/src/main/kotlin/conventions-root.gradle.kts
Mr3zee: It's just for our CI

PR #392, opened 2025-07-11
Support nullable contextual serializers

yakivy:
Context: https://kotlinlang.slack.com/archives/C072YJ3Q91V/p1752238186483329


Mr3zee (changes requested):
@yakivy Hey! Thanks for the contribution! I'll accept that, but can you please move the test into `krpc/krpc-test/src/commonMain/kotlin/kotlinx/rpc/krpc/test/KrpcTransportTestBase.kt`? Sampling is for the protocol testing purposes

You can see how `LocalDateSerializer` is used there for contextual serialization. Tests don't run in IDE, so you can check it with `./gradlew krpc:krpc-test:jvmTest --tests "*<test_name>*"`

Mr3zee:
@yakivy yep, thanks! One last thing: please put the test in the base class and just ignore it in protobuf. You can see how it's done for other nullable test

Mr3zee (approved):

Mr3zee:
@yakivy thanks for the contribution!

PR #397, opened 2025-07-15
Add stub targets tags for platforms table

Mr3zee:
**Subsystem**
Docs, Gradle

**Problem Description**
Sometimes (gRPC I'm looking at you) we have published targets, but they don't actually work and are just stubs.

**Solution**
Reflect these semantics in the platforms table.

**ATTENTION**
Provide this pull request with proper labels, it is important to simplify generation of releases.


e5l (approved):
lgtm

PR #411, opened 2025-07-25
Make protoc-gen-kotlinx (formerly protobuf-plugin) an included build

Mr3zee:
**Subsystem**
Protoc plugin

**Problem Description**
We want to test protoc plugin and we want to test gradle plugin.
But protoc plugin is now in the main project, and it imports the gradle plugin

**Solution**
So we put the protoc plugin as an included build


e5l (approved):
lgtm

PR #412, opened 2025-07-25
gRPC/Native: Add Wire Format encoder and decoder

Jozott00:
**Subsystem**
gRPC/Native

**Problem Description**
One part of the gRPC implementation is the Protobuf wire format encoding and decoding. 
This PR adds support for all values (except for sub-messages) to encode them into a `kotlinx.io.Buffer` and decode a buffer back to Kotlin values. 

**Solution**
This PR adds two new main interfaces to the `commonMain`: `WireEncoder` and `WireDecode`.
Both interfaces must be implemented platform specifically. This PR provides implementations for the Kotlin Native platform.

Most of the encoding and decoding is done by the Protobuf C++ library, which is called using C-Interop. To avoid unnecessary data copying during decoding, the C++ decoder is directly bridged to the `ZeroCopyInputSource` that provides low-level access to the `kotlinx.io.Buffer`. The implementation of the `ZeroCopyInputSource` highly relies on the implementation of `kotlinx.io.Buffer`.

### DIFF
--- a/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.kt
+++ b/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.kt
@@ -11,13 +11,11 @@ import kotlinx.io.Buffer
  *
  * This decoder is used by first calling [readTag], than looking up the field based on the field number in the returned,
  * tag and then calling the actual `read*()` method to read the value to the corresponding field.
- * This means that the nullable return value does not collide with optional fields, as optional fields would not
- * include a tag in the encoded message.
  *
- * If one `read*()` method returns `null`, decoding the data failed and no further
- * decoding can be done.
+ * [hadError] indicates an error during decoding. While calling `read*()` is safe, the returned values
+ * are meaningless if [hadError] returns `true`.
  *
- * NOTE: If the value of a `read*()` method is non-null, it doesn't mean that the
+ * NOTE: If the [hadError] after a call to `read*()` returns `false`, it doesn't mean that the
  * value is correctly decoded. E.g., the following test will pass:
  * ```kt
  * val fieldNr = 1
@@ -29,43 +27,45 @@ import kotlinx.io.Buffer
  *
  * WireDecoder(buffer).use { decoder ->
  *     decoder.readTag()
- *     assertNotNull(decoder.readBool())
+ *     decoder.readBool()
+ *     assertFalse(decoder.hasError())
  * }
  * ```
  */
 internal interface WireDecoder : AutoCloseable {
+    fun hadError(): Boolean
     fun readTag(): KTag?
-    fun readBool(): Boolean?
-    fun readInt32(): Int?
-    fun readInt64(): Long?
-    fun readUInt32(): UInt?
-    fun readUInt64(): ULong?
-    fun readSInt32(): Int?
-    fun readSInt64(): Long?
-    fun readFixed32(): UInt?
-    fun readFixed64(): ULong?
-    fun readSFixed32(): Int?
-    fun readSFixed64(): Long?
-    fun readFloat(): Float?
-    fun readDouble(): Double?
+    fun readBool(): Boolean
+    fun readInt32(): Int
+    fun readInt64(): Long
+    fun readUInt32(): UInt
+    fun readUInt64(): ULong
+    fun readSInt32(): Int
+    fun readSInt64(): Long
+    fun readFixed32(): UInt
+    fun readFixed64(): ULong
+    fun readSFixed32(): Int
+    fun readSFixed64(): Long
+    fun readFloat(): Float
+    fun readDouble(): Double
 
-    fun readEnum(): Int?
-    fun readString(): String?
-    fun readBytes(): ByteArray?
-    fun readPackedBool(): List<Boolean>?
-    fun readPackedInt32(): List<Int>?
-    fun readPackedInt64(): List<Long>?
-    fun readPackedSInt32(): List<Int>?
-    fun readPackedSInt64(): List<Long>?
-    fun readPackedUInt32(): List<UInt>?
-    fun readPackedUInt64(): List<ULong>?
-    fun readPackedFixed32(): List<UInt>?
-    fun readPackedFixed64(): List<ULong>?
-    fun readPackedSFixed32(): List<Int>?
-    fun readPackedSFixed64(): List<Long>?
-    fun readPackedFloat(): List<Float>?
-    fun readPackedDouble(): List<Double>?
-    fun readPackedEnum(): List<Int>?
+    fun readEnum(): Int
+    fun readString(): String
+    fun readBytes(): ByteArray
+    fun readPackedBool(): List<Boolean>
+    fun readPackedInt32(): List<Int>
+    fun readPackedInt64(): List<Long>
+    fun readPackedSInt32(): List<Int>
+    fun readPackedSInt64(): List<Long>
+    fun readPackedUInt32(): List<UInt>
+    fun readPackedUInt64(): List<ULong>
+    fun readPackedFixed32(): List<UInt>
+    fun readPackedFixed64(): List<ULong>
+    fun readPackedSFixed32(): List<Int>
+    fun readPackedSFixed64(): List<Long>
+    fun readPackedFloat(): List<Float>
+    fun readPackedDouble(): List<Double>
+    fun readPackedEnum(): List<Int>
 }
 
 /**
--- a/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
+++ b/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
@@ -19,6 +19,8 @@ private const val MAX_PACKED_BULK_SIZE: Int = 1_000_000
 @OptIn(ExperimentalForeignApi::class, ExperimentalNativeApi::class)
 internal class WireDecoderNative(private val source: Buffer) : WireDecoder {
 
+    private var hadError = false;
+
     // wraps the source in a class that allows to pass data from the source buffer to the C++ encoder
     // without copying it to an intermediate byte array.
     private val zeroCopyInput = StableRef.create(ZeroCopyInputSource(source))
@@ -61,149 +63,130 @@ internal class WireDecoderNative(private val source: Buffer) : WireDecoder {
         zeroCopyInput.dispose()
     }
 
+    override fun hadError(): Boolean {
+        return hadError;
+    }
+
     override fun readTag(): KTag? {
         val tag = pw_decoder_read_tag(raw)
-        return KTag.fromOrNull(tag)
+        if (tag == 0u) return null.withError()
+        val kTag = KTag.fromOrNull(tag)
+        if (kTag == null) {
+            hadError = true
+        }
+        return kTag
     }
 
-    override fun readBool(): Boolean? = memScoped {
+    override fun readBool(): Boolean = memScoped {
         val value = alloc<BooleanVar>()
-        if (pw_decoder_read_bool(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_bool(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readInt32(): Int? = memScoped {
+    override fun readInt32(): Int = memScoped {
         val value = alloc<IntVar>()
-        if (pw_decoder_read_int32(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_int32(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readInt64(): Long? = memScoped {
+    override fun readInt64(): Long = memScoped {
         val value = alloc<LongVar>()
-        if (pw_decoder_read_int64(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_int64(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readUInt32(): UInt? = memScoped {
+    override fun readUInt32(): UInt = memScoped {
         val value = alloc<UIntVar>()
-        if (pw_decoder_read_uint32(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_uint32(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readUInt64(): ULong? = memScoped {
+    override fun readUInt64(): ULong = memScoped {
         val value = alloc<ULongVar>()
-        if (pw_decoder_read_uint64(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_uint64(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readSInt32(): Int? = memScoped {
+    override fun readSInt32(): Int = memScoped {
         val value = alloc<IntVar>()
-        if (pw_decoder_read_sint32(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_sint32(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readSInt64(): Long? = memScoped {
+    override fun readSInt64(): Long = memScoped {
         val value = alloc<LongVar>()
-        if (pw_decoder_read_sint64(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_sint64(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readFixed32(): UInt? = memScoped {
+    override fun readFixed32(): UInt = memScoped {
         val value = alloc<UIntVar>()
-        if (pw_decoder_read_fixed32(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_fixed32(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readFixed64(): ULong? = memScoped {
+    override fun readFixed64(): ULong = memScoped {
         val value = alloc<ULongVar>()
-        if (pw_decoder_read_fixed64(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_fixed64(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readSFixed32(): Int? = memScoped {
+    override fun readSFixed32(): Int = memScoped {
         val value = alloc<IntVar>()
-        if (pw_decoder_read_sfixed32(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_sfixed32(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readSFixed64(): Long? = memScoped {
+    override fun readSFixed64(): Long = memScoped {
         val value = alloc<LongVar>()
-        if (pw_decoder_read_sfixed64(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_sfixed64(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readFloat(): Float? = memScoped {
+    override fun readFloat(): Float = memScoped {
         val value = alloc<FloatVar>()
-        if (pw_decoder_read_float(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_float(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readDouble(): Double? = memScoped {
+    override fun readDouble(): Double = memScoped {
         val value = alloc<DoubleVar>()
-        if (pw_decoder_read_double(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_double(raw, value.ptr).checkError()
+        return value.value
     }
 
-    override fun readEnum(): Int? = memScoped {
+    override fun readEnum(): Int = memScoped {
         val value = alloc<IntVar>()
-        if (pw_decoder_read_enum(raw, value.ptr)) {
-            return value.value
-        }
-        return null
+        pw_decoder_read_enum(raw, value.ptr).checkError()
+        return value.value
     }
 
     // TODO: Is it possible to avoid copying the c_str, by directly allocating a K/N String (as in readBytes)? KRPC-187
-    override fun readString(): String? = memScoped {
+    override fun readString(): String = memScoped {
         val str = alloc<CPointerVar<pw_string_t>>()
-        val ok = pw_decoder_read_string(raw, str.ptr)
+        pw_decoder_read_string(raw, str.ptr).checkError()
         try {
-            if (!ok) return null
-            return pw_string_c_str(str.value)?.toKString()
+            if (hadError) return ""
+            return pw_string_c_str(str.value)?.toKString() ?: "".also { hadError = true }
         } finally {
             pw_string_delete(str.value)
         }
     }
 
     // TODO: Should readBytes return a buffer, to prevent allocation of large contiguous memory blocks ? KRPC-182
-    override fun readBytes(): ByteArray? {
-        val length = readInt32() ?: return null
-        if (length < 0) return null
+    override fun readBytes(): ByteArray {
+        val length = readInt32()
+        if (hadError) return ByteArray(0)
+        if (length < 0) return ByteArray(0).withError()
         // check if the remaining buffer size is less than the set length,
         // we can early abort, without allocating unnecessary memory
-        if (source.size < length) return null
-        if (length == 0) return ByteArray(0)
+        if (source.size < length) return ByteArray(0).withError()
+        if (length == 0) return ByteArray(0) // actually an empty array (no error)
         val bytes = ByteArray(length)
-        var ok = true
         bytes.usePinned {
-            ok = pw_decoder_read_raw_bytes(raw, it.addressOf(0), length)
+            pw_decoder_read_raw_bytes(raw, it.addressOf(0), length).checkError()
         }
-        if (!ok) return null
+        if (hadError) return ByteArray(0)
         return bytes
     }
 
@@ -259,19 +242,21 @@ internal class WireDecoderNative(private val source: Buffer) : WireDecoder {
     )
 
     private inline fun <T : Any> readPackedVarInternal(
-        crossinline readFn: () -> T?
-    ): List<T>? {
-        val byteLen = readInt32() ?: return null
-        if (byteLen < 0) return null
-        if (source.size < byteLen) return null
-        if (byteLen == 0) return emptyList()
+        crossinline readFn: () -> T
+    ): List<T> {
+        val byteLen = readInt32()
+        if (hadError) return emptyList()
+        if (byteLen < 0) return emptyList<T>().withError()
+        if (source.size < byteLen) return emptyList<T>().withError()
+        if (byteLen == 0) return emptyList() // actually an empty list (no error)
 
         val limit = pw_decoder_push_limit(raw, byteLen)
 
         val result = mutableListOf<T>()
 
         while (pw_decoder_bytes_until_limit(raw) > 0) {
-            val elem = readFn() ?: return null
+            val elem = readFn()
+            if (hadError) break
             result.add(elem)
         }
 
@@ -293,13 +278,14 @@ internal class WireDecoderNative(private val source: Buffer) : WireDecoder {
         crossinline createArray: (Int) -> R,
         crossinline getAddress: Pinned<R>.(Int) -> COpaquePointer,
         crossinline asList: (R) -> List<T>
-    ): List<T>? {
+    ): List<T> {
         // fetch the size of the packed repeated field
-        var byteLen = readInt32() ?: return null
-        if (byteLen < 0) return null
-        if (source.size < byteLen) return null
-        if (byteLen % sizeBytes != 0) return null
-        if (byteLen == 0) return emptyList()
+        var byteLen = readInt32()
+        if (hadError) return emptyList()
+        if (byteLen < 0) return emptyList<T>().withError()
+        if (source.size < byteLen) return emptyList<T>().withError()
+        if (byteLen % sizeBytes != 0) return emptyList<T>().withError()
+        if (byteLen == 0) return emptyList()  // actually an empty list (no error)
 
         // allocate the buffer array (has at most MAX_PACKED_BULK_SIZE bytes)
         val bufByteLen = minOf(byteLen, MAX_PACKED_BULK_SIZE)
@@ -311,7 +297,7 @@ internal class WireDecoderNative(private val source: Buffer) : WireDecoder {
 
             if (byteLen == bufByteLen) {
                 // the whole packed field fits into the buffer -> copy into buffer and returns it as a list.
-                pw_decoder_read_raw_bytes(raw, bufAddr, byteLen)
+                pw_decoder_read_raw_bytes(raw, bufAddr, byteLen).checkError()
                 return asList(buffer)
             } else {
                 // the packed field is too large for the buffer, so we load it into a persistent list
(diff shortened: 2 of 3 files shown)


grpc/grpcpp-c/include/protowire.h
```diff
@@ -0,0 +1,136 @@
...
+    bool pw_decoder_read_double(pw_decoder_t *self, double *value);
+    bool pw_decoder_read_enum(pw_decoder_t *self, int *value);
+    bool pw_decoder_read_string(pw_decoder_t *self, pw_string_t **opaque_string);
+    // To read an actual bytes field, you must combine read_int32 and this function
+    bool pw_decoder_read_raw_bytes(pw_decoder_t *self, void* buffer, int size);
+
+    int pw_decoder_push_limit(pw_decoder_t *self, int limit);
+    void pw_decoder_pop_limit(pw_decoder_t *self, int limit);
+    int pw_decoder_bytes_until_limit(pw_decoder_t *self);
+
```
Mr3zee: Can you please add a description to these methods?

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+}
+
+struct pw_string {
+    std::string str;
+};
+
+struct pw_encoder {
+    protowire::SinkStream sinkStream;
+    pb::io::CopyingOutputStreamAdaptor cosa;
+    pb::io::CodedOutputStream cos;
```
Mr3zee: Let's give those two full names, instead of cosa/cos

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+    public:
+        /**
+         * Constructs the stream with a ctx pointer and a callback to the K/N side.
+         * The ctx pointer is used to on the K/N to reference a Kotlin managed object
+         * from within its static callback function.
+         *
+         * @param ctx the context used by the K/N side to reference Kotlin managed objects.
+         * @param sink the K/N callback to write data into the sink
+         */
+        SinkStream(void *ctx, bool(*sink)(void *ctx, const void *buffer, int size))
```
Mr3zee: > The ctx pointer is used to on the K/N to reference a Kotlin managed object
from within its static callback function.

Can we in that case give it a more explicit name? Like `objectReference` or `thisReference` or `thisObject`? I suspect that `this` is reserved

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+      cosa(&sinkStream),
+      cos(&cosa) {
+        cos.EnableAliasing(true);
+    }
+};
+
+struct pw_decoder {
+    protowire::BufferSourceStream ss;
+    pb::io::CodedInputStream cis;
+
```
Mr3zee: Same here with full names

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+              sink(sink) {
+        }
+
+        bool Write(const void *buffer, int size) override {
+            return sink(ctx, buffer, size);
+        }
+
+    private:
+        void *ctx;
+        bool (*sink)(void *ctx, const void *buffer, int size);
```
Mr3zee: You used `sink_fn` later in code, let's stick to it (`_fn` suffix) in all lambda parameter names

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+    static bool check(pw_encoder_t *self) {
+        return !self->cos.HadError();
+    }
+
+#define WRITE_FIELD_FUNC( funcSuffix, wireTy, cTy) \
+    bool pw_encoder_write_##funcSuffix(pw_encoder_t *self, int field_no, cTy value) { \
+        WireFormatLite::Write##wireTy(field_no, value, &self->cos); \
+        return check(self); \
+    }
+
```
Mr3zee: God, I haven't seen macros in years.. No other comments here 😁

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+    WRITE_FIELD_FUNC( fixed64, Fixed64, uint64_t)
+    WRITE_FIELD_FUNC( sfixed32, SFixed32, int32_t)
+    WRITE_FIELD_FUNC( sfixed64, SFixed64, int64_t)
+    WRITE_FIELD_FUNC( float, Float, float)
+    WRITE_FIELD_FUNC( double, Double, double)
+    WRITE_FIELD_FUNC( enum, Enum, int)
+
+    bool pw_encoder_write_string(pw_encoder_t *self, int field_no, const char *data, int size) {
+        return pw_encoder_write_bytes(self, field_no, data, size);
+    }
```
Mr3zee, line 179: Can we have any Charset problems here?

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+    bool pw_encoder_write_bytes(pw_encoder_t *self, int field_no, const void *data, int size) {
+        WireFormatLite::WriteTag(field_no, WireFormatLite::WIRETYPE_LENGTH_DELIMITED, &self->cos);
+        self->cos.WriteVarint32(size);
+        self->cos.WriteRawMaybeAliased(data, size);
+        return check(self);
+    }
+
+    bool pw_encoder_write_tag(pw_encoder_t *self, uint32_t tag) {
+        self->cos.WriteTag(tag);
```
Mr3zee: Line 177 and this one use different approaches to write tags, is it intended?

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+        return check(self);
+    }
+
+#define WRITE_FIELD_NO_TAG( funcSuffix, wireTy, cTy) \
+bool pw_encoder_write_##funcSuffix##_no_tag(pw_encoder_t *self, cTy value) { \
+WireFormatLite::Write##wireTy##NoTag(value, &self->cos); \
+return check(self); \
+}
```
Mr3zee: Probably line tabs are missing

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+    WRITE_FIELD_NO_TAG( fixed32, Fixed32, uint32_t)
+    WRITE_FIELD_NO_TAG( fixed64, Fixed64, uint64_t)
+    WRITE_FIELD_NO_TAG( sfixed32, SFixed32, int32_t)
+    WRITE_FIELD_NO_TAG( sfixed64, SFixed64, int64_t)
+    WRITE_FIELD_NO_TAG( float, Float, float)
+    WRITE_FIELD_NO_TAG( double, Double, double)
+    WRITE_FIELD_NO_TAG( enum, Enum, int)
+
+
+    /// DECODER IMPLEMENATION ///
```
Mr3zee: Please also add `ENCODER IMPLEMENATION` above

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+    void pw_decoder_delete(pw_decoder_t *self) {
+        delete self;
+    }
+
+    void pw_decoder_close(pw_decoder_t *self) {
+        // the deconstructor backs the stream up to the current position.
+        self->cis.~CodedInputStream();
+    }
+
```
Mr3zee: Do we close `zero_copy_input` somehow?

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+    READ_VAL_FUNC( sfixed64, SFIXED64, int64_t)
+    READ_VAL_FUNC( float, FLOAT, float)
+    READ_VAL_FUNC( double, DOUBLE, double)
+    READ_VAL_FUNC( enum, ENUM, int)
+
+    bool pw_decoder_read_string(pw_decoder_t *self, pw_string_t **string_ref) {
+        *string_ref = new pw_string_t;
+        return WireFormatLite::ReadString(&self->cis, &(*string_ref)->str);
+    }
+
```
Mr3zee: Same question about charsets here

grpc/grpcpp-c/src/protowire.cpp
```diff
@@ -0,0 +1,289 @@
...
+    uint32_t pw_size_uint64(uint64_t value) {
+     return WireFormatLite::UInt64Size(value);
+     }
+    uint32_t pw_size_sint32(int32_t value) {
+     return WireFormatLite::SInt32Size(value);
+     }
+    uint32_t pw_size_sint64(int64_t value) {
+     return WireFormatLite::SInt64Size(value);
+     }
+
```
Mr3zee: Please fix indents at this block

grpc/grpcpp-c/BUILD.bazel
```diff
@@ -8,14 +24,30 @@ cc_library(
     includes = ["include"],
     visibility = ["//visibility:public"],
     deps = [
+        # TODO: Reduce the dependencies and only use required once
```
Mr3zee: Please reference a YT ticket

grpc/grpcpp-c/BUILD.bazel
```diff
@@ -1,5 +1,21 @@
+load("@emsdk//emscripten_toolchain:wasm_rules.bzl", "wasm_cc_binary")
```
Mr3zee: Is it used somewhere now?

grpc/grpcpp-c/MODULE.bazel
```diff
@@ -22,3 +22,13 @@ bazel_dep(
...
+
+emsdk_version = "4.0.11"
+
+bazel_dep(name = "emsdk", version = emsdk_version)
+git_override(
+    module_name = "emsdk",
+    remote = "https://github.com/emscripten-core/emsdk.git",
+    strip_prefix = "bazel",
+    tag = emsdk_version,
+)
```
Mr3zee: What is this for?

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/KTag.kt
```diff
@@ -0,0 +1,51 @@
...
+        // Mask for those bits. (just 0b111)
+        val K_TAG_TYPE_MASK: UInt = (1u shl K_TAG_TYPE_BITS) - 1u
+    }
+}
+
+internal fun KTag.toRawKTag(): UInt {
+    return (fieldNr.toUInt() shl K_TAG_TYPE_BITS) or wireType.ordinal.toUInt()
+}
+
+internal fun KTag.Companion.from(rawKTag: UInt): KTag? {
```
Mr3zee: `fromOrNull`

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.kt
```diff
@@ -0,0 +1,75 @@
...
+    fun readFixed32(): UInt?
+    fun readFixed64(): ULong?
+    fun readSFixed32(): Int?
+    fun readSFixed64(): Long?
+    fun readFloat(): Float?
+    fun readDouble(): Double?
+    fun readEnum(): Int?
+    fun readString(): String?
+    fun readBytes(): ByteArray?
+    fun readPackedBool(): List<Boolean>?
```
Mr3zee: Do I get it right, that `packed` are the ones in a list (repeated field), and also for them we use write_no_tag (one of the previous comments by me)?

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.kt
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+package kotlinx.rpc.grpc.internal
+
+import kotlinx.io.Buffer
+
+/**
+ * A platform-specific decoder for wire format data.
+ *
+ * If one `read*()` method returns `null`, decoding the data failed and no further
```
Mr3zee: I'm pretty sure nullable types will be boxed. What about `hasBool`/`readBool` combo?

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.kt
```diff
@@ -0,0 +1,75 @@
...
+    fun readPackedSInt64(): List<Long>?
+    fun readPackedUInt32(): List<UInt>?
+    fun readPackedUInt64(): List<ULong>?
+    fun readPackedFixed32(): List<UInt>?
+    fun readPackedFixed64(): List<ULong>?
+    fun readPackedSFixed32(): List<Int>?
+    fun readPackedSFixed64(): List<Long>?
+    fun readPackedFloat(): List<Float>?
+    fun readPackedDouble(): List<Double>?
+    fun readPackedEnum(): List<Int>?
```
Mr3zee: Can we make enum more friendly in types here? If not - can you please add a how-to for encoding/decoding them?

grpc/grpc-core/src/jvmMain/java/kotlinx/rpc/grpc/internal/WireDecoder.jvm.kt
```diff
@@ -0,0 +1,5 @@
+/*
+ * Copyright 2023-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.rpc.grpc.internal
```
Mr3zee: I think actual declaration is missing here

grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Copyright 2023-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.rpc.grpc.internal
+
+import com.google.protobuf.CodedInputStream
+import com.google.protobuf.CodedOutputStream
+
+@OptIn(ExperimentalUnsignedTypes::class)
+internal class WireDecoder(val buffer: ByteArray) {
```
Mr3zee: Why this class is not in `WireDecoder.jvm.kt`?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/bufferUnsafeExtensions.kt
```diff
@@ -0,0 +1,30 @@
...
+import kotlinx.cinterop.*
+import kotlinx.io.InternalIoApi
+import kotlinx.io.Sink
+import kotlinx.io.UnsafeIoApi
+import kotlinx.io.unsafe.UnsafeBufferOperations
+import platform.posix.memcpy
+
+
+@OptIn(ExperimentalForeignApi::class, InternalIoApi::class, UnsafeIoApi::class)
+public fun Sink.writeFully(buffer: CPointer<ByteVar>, offset: Long, length: Long) {
```
Mr3zee: probably internal

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
```diff
@@ -0,0 +1,340 @@
...
+internal class WireDecoderNative(private val source: Buffer) : WireDecoder {
+
+    // wraps the source in a class that allows to pass data from the source buffer to the C++ encoder
+    // without copying it to an intermediate byte array.
+    private val zeroCopyInput = StableRef.create(ZeroCopyInputSource(source))
+
+    // construct the pw_decoder_t by passing a pw_zero_copy_input_t that provides a bridge between
+    // the CodedInputStream and the given source buffer. it passes functions that call the respective
+    // ZeroCopyInputSource methods.
+    internal val raw = run {
```
Mr3zee: Please add type, so It's easier to read it in PR 😅

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
```diff
@@ -0,0 +1,340 @@
...
+        // construct the pw_zero_copy_input_t that functions as a bridge to the ZeroCopyInputSource
+        val zeroCopyCInput = cValue<pw_zero_copy_input> {
+            ctx = zeroCopyInput.asCPointer()
+            next = staticCFunction { ctx, data, size ->
+                ctx!!.asStableRef<ZeroCopyInputSource>().get().next(data!!.reinterpret(), size!!.reinterpret())
+            }
```
Mr3zee, line 37: Can any of the parameters be null in any case?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
```diff
@@ -0,0 +1,340 @@
...
+        pw_decoder_new(zeroCopyCInput)
+            ?: error("Failed to create proto wire decoder")
+    }
+
+    val rawCleaner = createCleaner(raw) {
+        pw_decoder_delete(it)
+    }
+
+
+    override fun close() {
```
Mr3zee, line 57: should `rawCleaner` be disposed here somehow?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
```diff
@@ -0,0 +1,340 @@
...
+        zeroCopyInput.dispose()
+    }
+
+    override fun readTag(): KTag? {
+        val tag = pw_decoder_read_tag(raw)
+        return KTag.from(tag)
+    }
+
+    override fun readBool(): Boolean? = memScoped {
+        val value = alloc<BooleanVar>()
```
Mr3zee, line 81: What is `BooleanVar`?

Mr3zee: Is it boxed? If yes, can we mitigate it?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
```diff
@@ -0,0 +1,340 @@
...
+
+    override fun readEnum(): Int? = memScoped {
+        val value = alloc<IntVar>()
+        if (pw_decoder_read_enum(raw, value.ptr)) {
+            return value.value
+        }
+        return null
+    }
+
+    // TODO: Is it possible to avoid copying the c_str, by directly allocating a K/N String (as in readBytes)?
```
Mr3zee: Please add YT issue number

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
```diff
@@ -0,0 +1,340 @@
...
+        }
+    }
+
+    // TODO: Should readBytes return a buffer, to prevent allocation of large contiguous memory blocks ? KRPC-182
+    override fun readBytes(): ByteArray? {
+        val length = readInt32() ?: return null
+        if (length < 0) return null
+        // check if the remaining buffer size is less than the set length,
+        // we can early abort, without allocating unnecessary memory
+        if (source.size < length) return null
```
Mr3zee: Does `.size` return remaining size or whole?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
```diff
@@ -0,0 +1,340 @@
...
+        DoubleArray::asList,
+    )
+
+    private inline fun <T : Any> readPackedVarInternal(
+        crossinline readFn: () -> T?
+    ): List<T>? {
+        val byteLen = readInt32() ?: return null
+        if (byteLen < 0) return null
+        if (source.size < byteLen) return null
+        if (byteLen == 0) return null
```
Mr3zee: Why not `emptyList`?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
```diff
@@ -0,0 +1,340 @@
...
+import kotlinx.cinterop.*
+import kotlinx.collections.immutable.persistentListOf
+import kotlinx.io.Buffer
+import libprotowire.*
+import kotlin.experimental.ExperimentalNativeApi
+import kotlin.math.min
+import kotlin.native.ref.createCleaner
+
+// maximum buffer size to allocate as contiguous memory in bytes
+private const val MAX_PACKED_BULK_SIZE: Int = 1_000_000
```
Mr3zee, line 17: Please add an issue to benchmark different values for this

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    /*
+     * Based on the length of the packed repeated field, one of two list strategies is chosen.
+     * If the length is less or equal a specific threshold (MAX_PACKED_BULK_SIZE),
+     * a single array list is filled with the buffer-packed value (two copies).
+     * Otherwise, a kotlinx.collections.immutable.PersistentList is used to split allocation in several chunks.
+     * To build the persistent list, a buffer array is allocated that is used for fast copy from C++ to Kotlin.
+     *
+     * Note that this implementation assumes a little endian memory order.
```
Mr3zee, line 274: Does it metter if we receive a message from a client/server which is run on a big-endian host?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
```diff
@@ -0,0 +1,340 @@
...
+        val bufElemCount = bufByteLen / sizeBytes
+        val buffer = createArray(bufElemCount)
+
+        buffer.usePinned {
+            val bufAddr = it.getAddress(0)
+
+            if (byteLen == bufByteLen) {
+                // the whole packed field fits into the buffer -> copy into buffer and returns it as a list.
+                pw_decoder_read_raw_bytes(raw, bufAddr, byteLen)
+                return asList(buffer)
```
Mr3zee: So both copies are here right here, right? `read` and `asList`?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireEncoder.native.kt
```diff
@@ -0,0 +1,188 @@
...
+    /**
+     * The context object provides a stable reference to the kotlin context.
+     * This is required, as functions must be static and cannot capture environment references.
+     * With this context, the write callback (called by the pw_encoder_t) is able
+     * to write the data to the [sink].
+     */
+    private inner class Ctx {
+        fun write(buf: CPointer<ByteVar>, size: Int): Boolean {
+            sink.writeFully(buf, 0L, size.toLong())
+            return true
```
Mr3zee, line 25: Any case when it night be false?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireEncoder.native.kt
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    // create context as a stable reference that can be passed to static function callback
+    private val context = StableRef.create(this.Ctx())
+
+    // construct encoder with a callback that calls write() on this.context
+    internal val raw = run {
```
Mr3zee: Please add the type

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireEncoder.native.kt
```diff
@@ -0,0 +1,188 @@
...
+        }) ?: error("Failed to create proto wire encoder")
+    }
+
+    private val contextCleaner = createCleaner(context) {
+        it.dispose()
+    }
+    private val rawCleaner = createCleaner(raw) {
+        pw_encoder_delete(it)
+    }
+
```
Mr3zee, line 48: Are those cleaners used anywhere or can they be moved to an `init {}` block?

grpc/grpcpp-c/src/protowire.cpp
Jozott00: The Kotlin String to C string conversion (provided by C interop) uses utf-8 (which is also the supported charset of protobuf). So every Kotlin String is converted to a valid C string. So there should be no charset problem.

grpc/grpcpp-c/src/protowire.cpp
Jozott00: It used the underlying coded stream directly. I updated the API so that the user doesn't have to encode the field number and wire type by themselves, but instead pass both as argument.

grpc/grpcpp-c/src/protowire.cpp
Jozott00: The zero_copy_input is closed on the K/N side, as it's object is managed by Kotlin.

grpc/grpcpp-c/BUILD.bazel
Jozott00: Good catch, no it's currently not used. This was used to test the WASM compilation. I forgot to remove it

grpc/grpcpp-c/MODULE.bazel
Jozott00: Thanks for noting, I used it to test WASM compilation and forgot to remove it again.

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.kt
Jozott00: Right. Packed (repeated) fields share the same tag for across all values. Therefore, we only write one tag, followed by all values in the list (or read in this case). Note that for repeated (non-packed) fields, each value has its own tag.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireEncoder.native.kt
Jozott00: No, but the interface on the C++ side returns a boolean and we have to hardcode `return true` either here or on the C++ side. And I thought it would be more appropriate to do it here.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireEncoder.native.kt
Jozott00: Apparently, the GC calls the cleanup callback when collecting the cleaner object itself. So we have to bind the lifetime of the cleaner to the EncoderNative object, otherwise the GC might call the cleaner directly after the `init` block.

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.kt
Jozott00: You will only use a `read*()` method if you previously read the field with `readTag()`. And in this case, there must be a value. 

For optional proto fields, if it was not set in the message, it will never be read with `readTag()` so the `read*()` will never be called.

Jozott00: I will add a documentation on how to use this.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
Jozott00: No. The `ctx` is always passed to the callback as is, and `data` is non-null by definition of the C++ interface that declares the `next` method.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
Jozott00: AFAIK there is no API to cancel/dispose the `rawCleaner`. I separated the `close` and the `delete` of the encoder because the close is important to bring the source buffer into a valid state, while the delete must always be eventually called to prevent a memory leak. If I've moved the delete also into the close and the user doesn't close the decoder, we would end up in a memory leak. Also I cannot call delete in both, the cleaner and the `close` method, as this would lead to a double free crash.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
Jozott00: All C values are the type with the postfix `Var`. To get a modifiable pointer to some boolean value, I have to allocate it on the heap (apparently). So I think it is not possible to avoid the `alloc` call in this case.

Jozott00: I tried this:
```kt
override fun readInt32(): Int? = memScoped {
        val value = cValue<IntVar>()
        value.usePinned {
            if (pw_decoder_read_int32(raw, value.ptr)) {
                return value.ptr.pointed.value
            }
        }
        return null
    }
```
But it crashed

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
Jozott00: The size of that the buffer currently contains (between read pointer and write pointer)

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
Jozott00: Good catch! This should be an empty list!

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
Jozott00: No, as the protobuf wire format defines that the values are transferred in little endian.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.native.kt
Jozott00: The `asList` is actual not a copy, but only a `List` wrapper around the underlying array.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/WireEncoder.native.kt
Mr3zee: @Jozott00 In this case I saw some cleaners in init blocks in other places

Jozott00: Yes that is right, I will address it when I'm implementing the gRPC stub logic.

Mr3zee (approved):

PR #414, opened 2025-07-29
Added jvm codecs implementations

Mr3zee:
**Subsystem**
gRPC

**Problem Description**
We had codecs for K/N but not for jvm

**Solution**
Add them for jvm (and fix some issues along the way)

### DIFF
--- a/grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/internal/WireCodecTest.kt
+++ b/grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/internal/WireCodecTest.kt
@@ -4,13 +4,15 @@
 
 package kotlinx.rpc.grpc.internal
 
-import kotlinx.cinterop.ExperimentalForeignApi
 import kotlinx.io.Buffer
-import kotlin.experimental.ExperimentalNativeApi
 import kotlin.test.*
 
-// TODO: Move this to the commonTest
-@OptIn(ExperimentalForeignApi::class, ExperimentalNativeApi::class)
+enum class TestPlatform {
+    Jvm, Native, Js, WasmJs, Wasi;
+}
+
+expect val testPlatform: TestPlatform
+
 class WireCodecTest {
 
     @Test
@@ -343,8 +345,7 @@ class WireCodecTest {
         val buffer = Buffer()
 
         val decoder = WireDecoder(buffer)
-        decoder.readTag()
-        assertTrue(decoder.hadError())
+        assertNull(decoder.readTag())
     }
 
     @Test
@@ -418,6 +419,11 @@ class WireCodecTest {
 
     @Test
     fun testReadAfterClose() {
+        // jvm has no close method
+        if (testPlatform == TestPlatform.Jvm) {
+            return
+        }
+
         val fieldNr = 19
         val buffer = Buffer()
 
@@ -432,7 +438,7 @@ class WireCodecTest {
         try {
             val tag = decoder.readTag()
             assertNull(tag)
-        } catch (e: Exception) {
+        } catch (_: Exception) {
             // Expected exception in some implementations
         }
     }
@@ -497,6 +503,11 @@ class WireCodecTest {
 
     @Test
     fun testBufferNotExhausted() {
+        // jvm reads buffer - buffered (ba dum tss)
+        if (testPlatform == TestPlatform.Jvm) {
+            return
+        }
+
         val fieldNr = 1
         val buffer = Buffer()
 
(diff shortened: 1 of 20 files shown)


grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/internal/WireCodecTest.kt
```diff
@@ -497,6 +503,11 @@ class WireCodecTest {
 
     @Test
     fun testBufferNotExhausted() {
+        // jvm reads buffer - buffered (ba dum tss)
+        if (testPlatform == TestPlatform.Jvm) {
+            return
+        }
```
Jozott00, line 509: On JVM, what is the state of the buffer when the decoder doesn't read all bytes from the buffer/stream?

grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.jvm.kt
```diff
@@ -4,9 +4,123 @@
...
+import kotlinx.io.asInputStream
+
+internal class WireDecoderJvm(source: Buffer) : WireDecoder {
+    // there is no way to omit coping here
+    internal val codedInputStream: CodedInputStream = CodedInputStream.newInstance(source.asInputStream())
+
+    // errors in jvm are exceptions
+    override fun hadError(): Boolean {
+        return false
+    }
```
Jozott00, line 18: We should probably unify the behavior between all implementations. Probably, the K/N implementation should also throw exceptions instead of providing the error indicator? (in a separate PR)

grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.jvm.kt
```diff
@@ -4,9 +4,123 @@
...
+
+    override fun readTag(): KTag? {
+        codedInputStream
```
Jozott00: I think you forgot to remove this line.

grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.jvm.kt
```diff
@@ -4,9 +4,123 @@
...
+        return codedInputStream.readEnum()
+    }
+
+    override fun readString(): String {
+        return codedInputStream.readStringRequireUtf8()
+    }
+
+    override fun readBytes(): ByteArray {
+        // todo remove coping
+        return codedInputStream.readBytes().toByteArray()
```
Jozott00: I think there is also a `readBytesArray` method you could use to avoid coping.

grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/internal/WireEncoder.jvm.kt
```diff
@@ -4,8 +4,220 @@
...
+    }
+
+    override fun writeEnum(fieldNr: Int, value: Int): Boolean {
+        codedOutputStream.writeEnum(fieldNr, value)
+        return true
+    }
+
+    override fun writeBytes(fieldNr: Int, value: ByteArray): Boolean {
+        // todo remove coping
+        codedOutputStream.writeBytes(fieldNr, ByteString.copyFrom(value))
```
Jozott00: There is a `writeBytesArray` that takes a `byte[]` instead of a `ByteString`

grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.jvm.kt
Mr3zee: Yeah, let's do that in the next PR

grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.jvm.kt
Mr3zee: Thanks!

grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/internal/WireDecoder.jvm.kt
Mr3zee: Nice catch, thanks!

PR #418, opened 2025-07-31
protoc-plugin: Refactoring of intermediate model 

Jozott00:
**Subsystem**
Protoc Kotlin Common Plugin

**Problem Description**
Previously we created a Model directly from the `DescriptorProtos.*` objects. The problem with this approach is, that handling the semantics of multiple protobuf versions is hard.

**Solution**
Now, the model is build from the `Descriptor.*` objects that provide an abstraction over the raw proto descriptors. This allows us to generate protobuf version agnostic code. Additionally, message resolving is done by the descriptor, reducing the code logic on our side.


protoc-gen/src/main/kotlin/kotlinx/rpc/protobuf/RpcProtobufPlugin.kt
```diff
@@ -20,30 +20,9 @@ import java.io.File
 class RpcProtobufPlugin {
     companion object {
         private const val DEBUG_OUTPUT_OPTION = "debugOutput"
-        private const val MESSAGE_MODE_OPTION = "messageMode"
```
Mr3zee, line 23: Please also remove this option from the Gradle plugin

Mr3zee (approved):
Amazing! Please fix the imports (at least as they were before) and bring back the tests

PR #419, opened 2025-07-31
Grpc common services

Mr3zee:
**Subsystem**
gRPC Core

**Solution**
Common declarations for client and server


grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/MethodDescriptor.kt
```diff
@@ -0,0 +1,66 @@
...
+    public fun getSchemaDescriptor(): Any?
+    public fun isIdempotent(): Boolean
+    public fun isSafe(): Boolean
+    public fun isSampledToLocalTracing(): Boolean
+
+    public interface Marshaller<T> {
+        public fun stream(value: T): InputStream
+        public fun parse(stream: InputStream): T
+    }
+}
```
Jozott00, line 34: Why can't we define this as an interface instead of an `expect class`? 
EDIT: Ahh because you using it for a `actual typealias`

grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/pb/ProtosTest.kt
```diff
@@ -1,71 +1,3 @@
...
  */
-
-package kotlinx.rpc.grpc.pb
-
-import kotlinx.io.Buffer
-import kotlinx.rpc.grpc.test.common.*
-import kotlin.test.Test
-import kotlin.test.assertEquals
-
-class ProtosTest {
```
Jozott00: Did you remove this test class on purpose?

Jozott00 (approved):

grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/pb/ProtosTest.kt
Mr3zee: oops

PR #421, opened 2025-08-01
grpc-common: Add field presence tracking, required field enforcing and generation of the MessageCodec implementation

Jozott00:
**Subsystem**
gRPC/Common - Protobuf

**Description**
This PR adds presence tracking for fields with explicit presence on the wire. 
It does this by adding a `presenceMask` to each message, implemented as `BitSet`. The presence of each field that has explicit presence is represented by a bit in the mask. 

If a required field (`required` field in proto2 or a field annotated with `[ LEGACY_REQUIRED ]` in editions) is not present after construction, an error is thrown. 
This is done for both, user constructed messages (e.i. by the generated `invoke` function) and messages that were decoded from the wire using the message's decode function. 

Additionally, the PR adds generation of an implementation of the `MessageCodec` interface, used by the gRPC implementation to encode and decode messages. It is made available as object in the companion object of the generated internal message. E.g. `MyMessageInternal.CODEC`

### DIFF
--- a/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/InternalMessage.kt
+++ b/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/InternalMessage.kt
@@ -8,16 +8,7 @@ import kotlinx.rpc.grpc.utils.BitSet
 import kotlinx.rpc.internal.utils.InternalRpcApi
 
 @InternalRpcApi
-public abstract class Message(fieldsWithPresence: Int) {
-
+public abstract class InternalMessage(fieldsWithPresence: Int) {
     public val presenceMask: BitSet = BitSet(fieldsWithPresence)
 
-    @InternalRpcApi
-    public interface Companion<T : Message> {
-
-        public fun decodeWith(decoder: WireDecoder): T
-
-    }
-
-    public abstract fun encodeWith(encoder: WireEncoder)
 }
--- a/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/utils/BitSet.kt
+++ b/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/utils/BitSet.kt
@@ -2,11 +2,14 @@
  * Copyright 2023-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
  */
 
-package kotlinx.rpc.grpc.internal
+package kotlinx.rpc.grpc.utils
+
+import kotlinx.rpc.internal.utils.InternalRpcApi
 
 /**
  * A fixed-sized vector of bits, allowing one to set/clear/read bits from it by a bit index.
  */
+@InternalRpcApi
 public class BitSet(public val size: Int) {
     private val data: LongArray = LongArray((size + 63) ushr 6)
 
--- a/protoc-gen/src/main/kotlin/kotlinx/rpc/protobuf/ModelToKotlinCommonGenerator.kt
+++ b/protoc-gen/src/main/kotlin/kotlinx/rpc/protobuf/ModelToKotlinCommonGenerator.kt
@@ -105,7 +105,10 @@ class ModelToKotlinCommonGenerator(
 
         fileDeclaration.messageDeclarations.forEach {
             generateRequiredCheck(it)
+            generateMessageEncoder(it)
+            generateMessageDecoder(it)
         }
+
     }
 
     private fun MessageDeclaration.fields() = actualFields.map {
@@ -150,7 +153,7 @@ class ModelToKotlinCommonGenerator(
             declarationType = DeclarationType.Class,
             superTypes = listOf(
                 declaration.name.safeFullName(),
-                "Message(fieldsWithPresence = ${declaration.presenceMaskSize})"
+                "InternalMessage(fieldsWithPresence = ${declaration.presenceMaskSize})"
             ),
         ) {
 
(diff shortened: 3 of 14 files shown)


grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/InternalMessage.kt
```diff
@@ -0,0 +1,14 @@
...
+
+package kotlinx.rpc.grpc.pb
+
+import kotlinx.rpc.grpc.utils.BitSet
+import kotlinx.rpc.internal.utils.InternalRpcApi
+
+@InternalRpcApi
+public abstract class InternalMessage(fieldsWithPresence: Int) {
+    public val presenceMask: BitSet = BitSet(fieldsWithPresence)
+
```
Mr3zee: Please remove unnecessary NL

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/utils/BitSet.kt
```diff
@@ -0,0 +1,66 @@
...
+
+package kotlinx.rpc.grpc.utils
+
+import kotlinx.rpc.internal.utils.InternalRpcApi
+
+/**
+ * A fixed-sized vector of bits, allowing one to set/clear/read bits from it by a bit index.
+ */
+@InternalRpcApi
+public class BitSet(public val size: Int) {
```
Mr3zee, line 13: Maybe `sizeBits`?

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/utils/BitSet.kt
```diff
@@ -0,0 +1,66 @@
...
+    private val data: LongArray = LongArray((size + 63) ushr 6)
+
+    /** Sets the bit at [index] to 1. */
+    public fun set(index: Int) {
```
Mr3zee: Can also be `operator fun`

protoc-gen/src/main/kotlin/kotlinx/rpc/protobuf/ModelToKotlinCommonGenerator.kt
```diff
@@ -11,6 +11,8 @@ import kotlinx.rpc.protobuf.model.*
 import org.slf4j.Logger
 
 private const val RPC_INTERNAL_PACKAGE_SUFFIX = "_rpc_internal"
+private const val MSG_INTERNAL_SUFFIX = "Internal"
+private const val PB_PKG = "kotlinx.rpc.grpc.pb"
```
Mr3zee, line 15: This is for locating runtime APIs, right? Not to generate code

protoc-gen/src/main/kotlin/kotlinx/rpc/protobuf/ModelToKotlinCommonGenerator.kt
```diff
@@ -140,11 +145,19 @@ class ModelToKotlinCommonGenerator(
 
     @Suppress("detekt.CyclomaticComplexMethod")
     private fun CodeGenerator.generateInternalMessage(declaration: MessageDeclaration) {
+        val internalClassName = declaration.internalClassName()
         clazz(
-            name = "${declaration.name.simpleName}Builder",
+            name = internalClassName,
+            annotations = listOf("@kotlinx.rpc.internal.utils.InternalRpcApi"),
```
Mr3zee: Let's make it a const value (fqname)
Also let's add `@file:OptIn(<fqname>)` for the generated code

protoc-gen/src/main/kotlin/kotlinx/rpc/protobuf/ModelToKotlinCommonGenerator.kt
```diff
@@ -162,32 +175,84 @@ class ModelToKotlinCommonGenerator(
...
+                    code("const val ${field.name} = ${field.presenceIdx}")
+                    newLine()
+                }
+            }
+        }
+    }
+
+    private fun CodeGenerator.generateCodecObject(declaration: MessageDeclaration) {
+        val msgFqName = declaration.name.safeFullName()
+        val downCastErrorStr = "The message is a custom implementation of ${msgFqName}. This is currently not allowed."
```
Mr3zee: Didn't get the `This is currently not allowed.` from the first try.
Maybe something like `Class "this::name" implements ${msgFqName} which is prohibited`

protoc-gen/src/main/kotlin/kotlinx/rpc/protobuf/ModelToKotlinCommonGenerator.kt
```diff
@@ -162,32 +175,84 @@ class ModelToKotlinCommonGenerator(
...
+        scope("val CODEC = object : kotlinx.rpc.grpc.internal.MessageCodec<$msgFqName>") {
+            function("encode", modifiers = "override", args = "value: $msgFqName", returnType = "$sourceFqName") {
+                code("val msg = value as? ${declaration.internalClassFullName()} ?: error(\"$downCastErrorStr\")")
+                code("val buffer = $bufferFqName()")
+                code("val encoder = $PB_PKG.WireEncoder(buffer)")
+                code("msg.encodeWith(encoder)")
+                code("encoder.flush()")
+                code("return buffer")
+            }
+
```
Mr3zee, line 224: Please remove the extra NL

protoc-gen/src/main/kotlin/kotlinx/rpc/protobuf/ModelToKotlinCommonGenerator.kt
Jozott00: Right, in a previous version I imported `kotlinx.rpc.grpc.pb.*`. I thought it would be better to reference them with their fully qualified name.

Mr3zee: Yep, that's good

PR #427, opened 2025-08-04
Fix gRPC services tests

Mr3zee:
**Subsystem**
gRPC

**Problem Description**
Tests with streaming were not working

**Solution**
Closer rewrite the code and see that `Ready` has one more property for overall client/server readiness
Disclaimer: I have absolutely no idea why this fix works


grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/suspendClientCalls.kt
```diff
@@ -252,16 +252,22 @@ internal suspend fun <T> Flow<T>.singleOrStatus(
...
+                "Should be impossible; a CONFLATED channel should never return false on offer"
+            )
+        }
     }
 
     suspend fun suspendUntilReady() {
-        channel.receive()
+        while (!isReallyReady()) {
+            channel.receive()
+        }
```
Jozott00, line 271: I think there is still a case where `suspendUntilReady` won't return, even if the handler is ready. 

Following Scenario:
1. `Handler.isReady() == false`
2. Two coroutines call `suspendUntilReady` and both are suspend at `channel.receive()`
3. The handler turns ready, invokes onReady, a `Unit` is sent to the channel
4. One of the coroutines consumes the message and will return the `suspendUntilReady` function
5. The other one will still wait for an element on the channel. However, this might never be the case, as the handler might never turn `isReady` to `false` and then `true` and thus never calls `onReady` again.

I think we need some kind of notification/wakeUp mechanism that allows all currently waiting coroutines to continue.

Mr3zee: I think there is no scenario where two coroutines call the same ready object, so it should be fine

Mr3zee: Also, this is the exact same code as in grpc/kotlin😁

Jozott00: Ahh ok, alright, then it's fine ^^

PR #428, opened 2025-08-05
Added gRPC service generation to the compiler plugin

Mr3zee:
**Subsystem**
Compiler Plugin, gRPC

**Problem Description**
We have all platform bindings for the gRPC, but no service generation

**Solution**
Add service generations and a `MessageCodec` locator:
- Use `MessageCodecResolver` in `GrpcClient`/`GrpcServer`
- Or annotate your message with `WithCodec` annotation (default for protobuf messages)

And we now have **fully working gRPC** services with arbitrary serilization.

For next PRs:
- FIR checkers
- Method metadata

### DIFF
--- a/compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcStubGenerator.kt
+++ b/compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcStubGenerator.kt
@@ -922,8 +922,6 @@ internal class RpcStubGenerator(
                 visibility = DescriptorVisibilities.PUBLIC
                 modality = Modality.FINAL
             }.apply function@{
-                parameters = listOf(this@generateCallablesProperty.thisReceiver!!.copyTo(this))
-
                 val functionThisReceiver = vsApi {
                     stubCompanionObjectThisReceiver.copyToVS(this@function, origin = IrDeclarationOrigin.DEFINED)
                 }.also {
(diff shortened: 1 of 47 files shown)


grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/descriptor/GrpcServiceDescriptor.kt
```diff
@@ -4,29 +4,32 @@
...
+    public fun getMethodDescriptor(methodName: String): MethodDescriptor<*, *>? = methodDescriptorMap[methodName]
 }
 
-@ExperimentalRpcApi
-public interface GrpcClientDelegate {
-    public suspend fun <R> call(rpcCall: RpcCall): R
-
-    public fun <R> callServerStreaming(rpcCall: RpcCall): Flow<R>
-}
+//@ExperimentalRpcApi
```
e5l: please drop commented code

compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcStubGenerator.kt
```diff
@@ -1368,6 +1780,14 @@ internal class RpcStubGenerator(
         value = value,
     )
 
+    private fun IrType.unwrapFlow(): IrType {
```
e5l, line 1781: The class is huge. Please check what can be extracted as utility methods

e5l (approved):
please check comments

Jozott00 (approved):

compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcStubGenerator.kt
Mr3zee: I still think that one big file is better DX, for the current size at least.
And while I'm the only person making bug changes here, I'd like to have like this. I have in mind splitting it, but only when it's size becomes an issue during the development

PR #430, opened 2025-08-06
grpc: Support sub-messages

Jozott00:
**Subsystem**
gRPC/Protobuf

**Solution**
This PR adds support for sub-messages to the Protobuf generation.
Currently, messages are never nullable and there also no way to determine whether a sub-message field was set/present or not. [KRPC-188](https://youtrack.jetbrains.com/issue/KRPC-188)


grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/InternalMessage.kt
```diff
@@ -6,8 +6,41 @@ package kotlinx.rpc.grpc.pb
...
 import kotlinx.rpc.internal.utils.InternalRpcApi
+import kotlin.properties.ReadWriteProperty
+import kotlin.reflect.KProperty
 
 @InternalRpcApi
 public abstract class InternalMessage(fieldsWithPresence: Int) {
     public val presenceMask: BitSet = BitSet(fieldsWithPresence)
+
+    @Suppress("PropertyName")
+    public abstract val _size: Int
```
Mr3zee, line 17: Why is it with a `_` prefix?

grpc/grpc-core/src/commonTest/proto/nested.proto
```diff
@@ -0,0 +1,74 @@
+// Protocol Buffers - Google's data interchange format
+// Copyright 2023 Google LLC.  All rights reserved.
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
```
Mr3zee: Let's remove this

protoc-gen/src/main/kotlin/kotlinx/rpc/protobuf/ModelToKotlinCommonGenerator.kt
```diff
@@ -656,6 +870,9 @@ class ModelToKotlinCommonGenerator(
         }
     }
 
+    private fun FieldType.Message.internalCtor() =
```
Mr3zee: Let's make it `internalConstructor`

Mr3zee (approved):
Great speed

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/InternalMessage.kt
Jozott00: I thought `size` might be a very common field name, so I prefixed it with `_` to avoid naming conflicts. Do you know a better name/way to avoid conflicts?

Mr3zee: Oh, hmmmm 🤔

Mr3zee: Yep, no, that I didn't think of, `_size` is fine then

PR #431, opened 2025-08-06
Add collect once check for client streams

Mr3zee:
**Subsystem**
kRPC

**Problem Description**
Client flows have side effects and thus can't be collected more than once

### DIFF
--- a/krpc/krpc-server/src/commonMain/kotlin/kotlinx/rpc/krpc/server/internal/ServerStreamContext.kt
+++ b/krpc/krpc-server/src/commonMain/kotlin/kotlinx/rpc/krpc/server/internal/ServerStreamContext.kt
@@ -4,6 +4,7 @@
 
 package kotlinx.rpc.krpc.server.internal
 
+import kotlinx.atomicfu.atomic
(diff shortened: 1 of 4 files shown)


krpc/krpc-server/src/commonMain/kotlin/kotlinx/rpc/krpc/server/internal/ServerStreamContext.kt
```diff
@@ -92,5 +99,9 @@ internal class ServerStreamContext {
     private fun streamCanceled(): Throwable = NoSuchElementException("Stream canceled")
 }
 
+private class AtomicBoolean(initialValue: Boolean) {
```
e5l, line 102: 😁

e5l (approved):

krpc/krpc-server/src/commonMain/kotlin/kotlinx/rpc/krpc/server/internal/ServerStreamContext.kt
Mr3zee: 🌚

PR #437, opened 2025-08-08
grpc-pb: Add gRPC Exceptions and skip unknown fields

Jozott00:
**Subsystem**
gRPC/Protobuf

**Solution**
This PR adds the `GrpcException` as well its sealed sub types `ProtobufEncodingException` and `ProtobufDecodingException`. 

Additionally, unknown fields are now skipped without an error.

There was also a bug in the native `WireEncoder` that caused the string length to include the null termination character, resulting in an invalid protobuf wire encoding of the message.

### DIFF
--- a/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/WireDecoder.kt
+++ b/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/WireDecoder.kt
@@ -105,7 +105,7 @@ public interface WireDecoder : AutoCloseable {
  * Turns exceptions thrown by different platforms during decoding into [ProtobufDecodingException].
  */
 @InternalRpcApi
-public expect fun checkForPlatformDecodeException(block: () -> Unit)
+public expect inline fun checkForPlatformDecodeException(block: () -> Unit)
 
 /**
  * Creates a platform-specific [WireDecoder].
--- a/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/pb/WireDecoder.native.kt
+++ b/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/pb/WireDecoder.native.kt
@@ -302,6 +302,6 @@ internal class WireDecoderNative(private val source: Buffer) : WireDecoder {
 
 public actual fun WireDecoder(source: Buffer): WireDecoder = WireDecoderNative(source)
 
-public actual fun checkForPlatformDecodeException(block: () -> Unit) {
+public actual inline fun checkForPlatformDecodeException(block: () -> Unit) {
     block()
 }
--- a/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/ProtobufException.kt
+++ b/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/ProtobufException.kt
@@ -10,7 +10,7 @@ public sealed class ProtobufException : RuntimeException {
 
 
 public class ProtobufDecodingException : ProtobufException {
-    internal constructor(message: String, cause: Throwable? = null) : super(message, cause)
+    public constructor(message: String, cause: Throwable? = null) : super(message, cause)
 
     public companion object Companion {
         internal fun missingRequiredField(messageName: String, fieldName: String) =
@@ -36,5 +36,5 @@ public class ProtobufDecodingException : ProtobufException {
 }
 
 public class ProtobufEncodingException : ProtobufException {
-    internal constructor(message: String, cause: Throwable? = null) : super(message, cause)
+    public constructor(message: String, cause: Throwable? = null) : super(message, cause)
 }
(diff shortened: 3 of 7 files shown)


grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/KTag.kt
```diff
@@ -36,14 +36,14 @@ internal fun KTag.toRawKTag(): UInt {
     return (fieldNr.toUInt() shl KTag.Companion.K_TAG_TYPE_BITS) or wireType.ordinal.toUInt()
 }
 
-internal fun KTag.Companion.fromOrNull(rawKTag: UInt): KTag? {
+internal fun KTag.Companion.from(rawKTag: UInt): KTag {
     val type = (rawKTag and K_TAG_TYPE_MASK).toInt()
     val field = (rawKTag shr K_TAG_TYPE_BITS).toInt()
     if (!isValidFieldNr(field)) {
-        return null
+        error("Invalid field number: $field")
```
Mr3zee: Why not `ProtobufDecodingException`?

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/KTag.kt
```diff
@@ -36,14 +36,14 @@ internal fun KTag.toRawKTag(): UInt {
...
     }
     if (type >= WireType.entries.size) {
-        return null
+        error("Invalid wire type: $type")
```
Mr3zee: Same here

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/WireDecoder.kt
```diff
@@ -79,15 +78,24 @@ public interface WireDecoder : AutoCloseable {
...
         popLimit(limit)
     }
+
+    public fun skipValue(writeType: WireType) {
+        when (writeType) {
+            WireType.VARINT -> readInt64()
+            WireType.FIXED32 -> readFixed32()
+            WireType.FIXED64 -> readFixed64()
+            WireType.LENGTH_DELIMITED -> readBytes()
+            WireType.START_GROUP -> error("Unexpected START_GROUP wire type (KRPC-193)")
```
Mr3zee: Same here

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcException.kt
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2023-2025 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.rpc.grpc
+
+public sealed class GrpcException : RuntimeException {
```
Mr3zee: Let's make it ProtobufException, I want to decouple protobuf from grpc

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/KTag.kt
Jozott00: ~The check for an invalid number is already done at this point. So if `from` is called with an invalid tag, it is actually an implementation error on our side.~

I will turn it into a `ProtobufDecodingException`.

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/KTag.kt
Jozott00: I will turn it into an `ProtobufDecodingException`

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/pb/WireDecoder.kt
Jozott00: I was not sure if this is a `ProtobufDecodingException` as it is not an illegal wire type, just kind of a "bug" on our side. 

But yeah, I will turn it into a `ProtobufDecodingException` so it is easier to handle.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/pb/WireDecoder.native.kt
```diff
@@ -302,13 +296,12 @@ internal class WireDecoderNative(private val source: Buffer) : WireDecoder {
...
-        hadError = true
-        return this
+        if (!this) throw ProtobufDecodingException.genericParsingError()
     }
 }
 
-internal actual fun WireDecoder(source: Buffer): WireDecoder = WireDecoderNative(source)
+public actual fun WireDecoder(source: Buffer): WireDecoder = WireDecoderNative(source)
+
+public actual fun checkForPlatformDecodeException(block: () -> Unit) {
```
Mr3zee: Let's make it inline

PR #438, opened 2025-08-08
Use compat-patrouille for compatibility settings

Mr3zee:
**Subsystem**
Gradle

There is a tool that configures all compatibility related flags for JVM and Kotlin:
https://github.com/GradleUp/compat-patrouille

**Solution**
We might as well use it


versions-root/libs.versions.toml
```diff
@@ -28,6 +27,7 @@ detekt-gradle-plugin = "1.23.8"
 kover = "0.9.1"
 develocity = "3.19.2"
 common-custom-user-data = "2.3"
+compat-patrouille = "0.0.1"
```
e5l, line 30: 🐾

e5l (approved):

PR #443, opened 2025-08-11
Grpc protobuf split

Mr3zee:
**Subsystem**
gRPC, Protoc-gen

**Problem Description**
We initially merged gRPC and protobuf declarations, however, they are not coupled

**Solution**
Extract protobuf into a separate module and split protoc-gen into two


Mr3zee:
@Jozott00 for some reason I get 
```
Exception in thread "main" java.lang.Error: /Users/Alexander.Sysoev/Desktop/AlexFolder/JetBrains/kotlinx-rpc/cinterop-c/include/grpcpp_c.h:9:10: fatal error: 'grpc/slice.h' file not found
```
Do you how to deal with it?

PR #445, opened 2025-08-12
fix wrong unchecked null cast (potential NPE)

y9maly:
**Problem Description**

I encountered an issue where an RPC service returned null even though the RPC function never actually returns null

After investigation, it turned out that the error occurred during the deserialization of the server's response. However, kotlinx-rpc silently ignored the deserialization exception and performed an unchecked cast of null to the RPC function's return type T

This happens inside the KrpcClient class in the handleServerStreamingMessage function:

```Kotlin
is KrpcCallMessage.CallSuccess, is KrpcCallMessage.StreamMessage -> {
    val value = runCatching {
        val serializerResult = serialFormat.serializersModule
            .buildContextual(callable.returnType)
        decodeMessageData(serialFormat, serializerResult, message)
    }
    @Suppress("UNCHECKED_CAST")
    channel.send(value.getOrNull() as T) // Thats not ok
}
```

When deserialization fails, getOrNull() returns null, which is then unchecked-cast to T. This may propagate unexpected null values

**Expected behaviour**

If deserialization fails, the error should be propagated back to the remote function call site, instead of sending a null value through the channel

Replacing getOrNull() with getOrThrow() would throw the exception immediately inside the library's internals, but this would not deliver it to the location where the developer actually invokes the remote function.
Therefore, I switched from Channel<T> to Channel<Result<T>>, allowing the deserialization error to be captured and then rethrown at the correct call site

This is my first attempt at contributing, please don’t hit me 🙃

### DIFF
--- a/krpc/krpc-test/src/commonMain/kotlin/kotlinx/rpc/krpc/test/KrpcTransportTestBase.kt
+++ b/krpc/krpc-test/src/commonMain/kotlin/kotlinx/rpc/krpc/test/KrpcTransportTestBase.kt
@@ -20,6 +20,7 @@ import kotlinx.rpc.krpc.server.KrpcServer
 import kotlinx.rpc.registerService
 import kotlinx.rpc.withService
 import kotlinx.serialization.KSerializer
+import kotlinx.serialization.SerializationException
 import kotlinx.serialization.descriptors.PrimitiveKind
 import kotlinx.serialization.descriptors.PrimitiveSerialDescriptor
 import kotlinx.serialization.descriptors.SerialDescriptor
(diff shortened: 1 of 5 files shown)


Mr3zee:
Hey @y9maly! Thanks for you contribution! I'll take a look at it tomorrow

y9maly:
@Mr3zee i did it

krpc/krpc-test/src/commonMain/kotlin/kotlinx/rpc/krpc/test/KrpcTransportTestBase.kt
```diff
@@ -460,6 +461,16 @@ abstract class KrpcTransportTestBase {
...
         assertEquals(Unit, client.unitFlow().toList().single())
     }
+
+    @Test
+    fun testPR445() = runTest {
+        assertFailsWith<SerializationException> {
+            val result = client.returnTestClassThatThrowsWhileDeserialization(42)
+            @Suppress("SENSELESS_COMPARISON")
+            if (result == null)
+                println("result must not be null")
```
Mr3zee:
```suggestion
            if (result == null) {
                assertNotNull(result, "result must not be null")
            }
```

Mr3zee (changes requested):
@y9maly thanks! One small change, please, while I'll run the tests

Mr3zee:
Our CI is broken, so I ran tests locally, and everything seams good.
Thank you for your great contribution!

Mr3zee (approved):

Mr3zee:
Will be released in 0.10.0 in upcoming weeks

PR #452, opened 2025-08-18
grpc-native: Support client calls on native

Jozott00:
**Subsystem**
gRPC/Native

**Solution**
This PR uses the grpc-core (C) library to support client calls for Kotin/Native.


grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/test/RawClientTest.kt
```diff
@@ -0,0 +1,110 @@
...
+            block(channel.platformApi, methodDescriptor)
+        } finally {
+            channel.shutdown()
+            channel.awaitTermination()
+        }
+
+
+    }
+
+
```
Mr3zee: Please remove unnecessary line breaks

grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/test/RawClientTest.kt
```diff
@@ -0,0 +1,110 @@
...
+            assertEquals("Eccchhooo", it.message)
+        }
+        assertEquals(5, i)
+    }
+
+    fun runTest(
+        methodName: String,
+        type: MethodType,
+        block: suspend (GrpcChannel, MethodDescriptor<EchoRequest, EchoResponse>) -> Unit,
+    ) = runTest {
```
Mr3zee, line 83: This test will fail without a server
Please either add a todo or a setup to run a test server

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/CallbackFuture.kt
```diff
@@ -0,0 +1,49 @@
...
+package kotlinx.rpc.grpc.internal
+
+import kotlinx.atomicfu.atomic
+
+/**
+ * Thread safe future for callbacks.
+ */
+internal class CallbackFuture<T : Any> {
+    private sealed interface State<out T> {
+        data class Pending<T>(val cbs: List<(T) -> Unit> = emptyList()) : State<T>
```
Mr3zee: Please name `callbacks`

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/CompletionQueue.kt
```diff
@@ -0,0 +1,193 @@
...
+     * - the queue is closed
+     * - the queue is in the process of a force shutdown
+     * - the queue is in the process of a normal shutdown, and the batch is a new `RECV_STATUS_ON_CLIENT` batch.
+     */
+    object CQShutdown : BatchResult
+
+    /**
+     * Happens when the batch couldn't be submitted for some reason.
+     */
+    data class CallError(val error: grpc_call_error) : BatchResult
```
Mr3zee: Let's decide on either `Called` or `Submitted` naming, so that docs and API are consistent

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/CompletionQueue.kt
```diff
@@ -0,0 +1,193 @@
...
+ * the queue.
+ * Users can attach to the returned [CallbackFuture] if the batch was successfully submitted (see [BatchResult]).
+ */
+internal class CompletionQueue {
+
+    internal enum class State { OPEN, SHUTTING_DOWN, CLOSED }
+
+    // if the queue was called with forceShutdown = true,
+    // it will reject all new batches and wait for all current ones to finish.
+    private var forceShutdown = false
```
Mr3zee: Is `CompletionQueue` thread-safe? Please, add the mention of it in the docs if not

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/CompletionQueue.kt
```diff
@@ -0,0 +1,193 @@
...
+ * It is based on the "new" callback API; therefore, there are no kotlin-side threads required to poll
+ * the queue.
+ * Users can attach to the returned [CallbackFuture] if the batch was successfully submitted (see [BatchResult]).
+ */
+internal class CompletionQueue {
+
+    internal enum class State { OPEN, SHUTTING_DOWN, CLOSED }
+
+    // if the queue was called with forceShutdown = true,
+    // it will reject all new batches and wait for all current ones to finish.
```
Mr3zee: Is this the terminology from the C lib? It's a bit confusing, as to I wouldn't expect forced shutdown to wait for anything. GrpcServer/Client have a difference of shutdown/shutdownNow for this, is there anything like this here?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/CompletionQueue.kt
```diff
@@ -0,0 +1,193 @@
...
+
+        return BatchResult.Called(completion)
+    }
+
+    /**
+     * Shuts down the queue.
+     * The method returns immediately, but the queue will be shut down asynchronously.
+     * The returned [CallbackFuture] will be completed with `Unit` when the queue is shut down.
+     */
+    fun shutdown(force: Boolean = false): CallbackFuture<Unit> {
```
Mr3zee, line 140: Please describe the effect of the `force` parameter here

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/CompletionQueue.kt
```diff
@@ -0,0 +1,193 @@
...
+        // however, this lock will be available very fast, so it shouldn't be a problem.'
+        synchronized(batchStartGuard) {
+            grpc_completion_queue_shutdown(raw)
+        }
+
+        return _shutdownDone
+    }
+}
+
+// kq stands for kompletion_queue lol
```
Mr3zee, line 161: 😄

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -0,0 +1,383 @@
...
+                    cancelInternal(grpc_status_code.GRPC_STATUS_INTERNAL, "Call failed: ${it.message}")
+                }
+            }
+        }
+    }
+
+    private var listener: Listener<Response>? = null
+    private var halfClosed = false
+    private var cancelled = false
+    private var closed = atomic(false)
```
Mr3zee, line 54: `NativeClientCall` shouldn't be thread safe, right?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -0,0 +1,383 @@
...
+
+    // tracks how many operations are in flight (not yet completed by the listener).
+    // if 0, there are no more operations (except for the RECV_STATUS_ON_CLIENT op).
```
Mr3zee: Can there be a situation when new OP comes after `inFlight` hit `0`?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -0,0 +1,383 @@
...
+    // we need this mechanism to ensure that onClose is not called while any other callback is still running
+    // on the listener.
+    private val inFlight = atomic(0)
+
+    // holds the received status information returned by the RECV_STATUS_ON_CLIENT batch.
+    // if null, the call is still in progress. otherwise, the call can be closed as soon as inFlight is 0.
+    private val closeInfo = atomic<Pair<Status, GrpcTrailers>?>(null)
+
+    // we currently don't buffer messages, so after one `sendMessage` call, ready turns false.
+    private val ready = atomic(true)
```
Mr3zee: Is it hard (or does it make sense) to buffer them?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -0,0 +1,383 @@
...
+    /**
+     * Tries to close the call by invoking the listener's onClose callback.
+     *
+     * - If the call is already closed, this does nothing.
+     * - If the RECV_STATUS_ON_CLIENT batch is still in progress, this does nothing.
+     * - If the [inFlight] counter is not 0, this does nothing.
+     * - Otherwise, the listener's onClose callback is invoked and the call is closed.
+     */
+    private fun tryDeliverClose() {
+        val s = closeInfo.value ?: return
```
Mr3zee: please add a meaningful name

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -0,0 +1,383 @@
...
+        if (inFlight.value == 0 && closed.compareAndSet(expect = false, update = true)) {
```
Mr3zee, line 102: `if (inFlight.value == 0`

Can `inFlight` go from `0` and up? If yes, this is a race condition here

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -0,0 +1,383 @@
...
+            val lst = checkNotNull(listener) { "Not yet started" }
+            // allows the managed channel to join for the call to finish.
+            callJob.complete()
+            lst.onClose(s.first, s.second)
+        }
+    }
+
+    /**
+     * Sets the [closeInfo] and calls [tryDeliverClose].
+     * This is called as soon as the RECV_STATUS_ON_CLIENT batch is finished.
```
Mr3zee: Where can I read more about `RECV_STATUS_ON_CLIENT`, I'm a little lost on what it means

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -0,0 +1,383 @@
...
+
+        when (val callResult = cq.runBatch(this@NativeClientCall, ops, nOps)) {
+            is BatchResult.Called -> {
+                callResult.future.onComplete { success ->
+                    try {
+                        if (success) {
+                            onSuccess()
+                        }
+                    } finally {
+                        // ignore failure, as it is reflected in the client status op
```
Mr3zee, line 174: Just to check, it there a test for it?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -0,0 +1,383 @@
...
+        ops: CPointer<grpc_op>,
+        nOps: ULong,
+        cleanup: () -> Unit = {},
+        onSuccess: () -> Unit = {},
+    ) {
+        // we must not try to run a batch after the call is closed.
+        if (closed.value) return cleanup()
+
+        // pre-book the batch, so onClose cannot be called before the batch finished.
+        beginOp()
```
Mr3zee, line 164: What if nOps is more than 1?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -0,0 +1,383 @@
...
+    override fun request(numMessages: Int) {
+        check(numMessages > 0) { "numMessages must be > 0" }
+        val listener = checkNotNull(listener) { "Not yet started" }
+        check(!cancelled) { "Already cancelled" }
+
+        var remainingMessages = numMessages
+
+        // we need to request only one message at a time, so we use a recursive function that
+        // requests one message and then calls itself again.
+        fun post() {
```
Mr3zee, line 298: Looks like a potential stackoverflow

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -0,0 +1,383 @@
...
+        }
+    }
+
+    override fun isReady(): Boolean = ready.value
+
+    override fun sendMessage(message: Request) {
+        checkNotNull(listener) { "Not yet started" }
+        check(!halfClosed) { "Already half closed." }
+        check(!cancelled) { "Already cancelled." }
+        check(isReady()) { "Not yet ready." }
```
Mr3zee: Please make the checks in this and other places something like, `internalError("Already half closed")` and `fun internalError(message: String) { error("Unexpected internal error: $message. Please, report the issue here: https://github.com/Kotlin/kotlinx-rpc/issues/new?template=bug_report.md")}`

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -0,0 +1,383 @@
...
+        val byteBuffer = inputStream.asSource().toGrpcByteBuffer()
+
+        val op = arena.alloc<grpc_op> {
+            op = GRPC_OP_SEND_MESSAGE
+            data.send_message.send_message = byteBuffer
+        }
+
+        runBatch(op.ptr, 1u, cleanup = {
+            // no mater what happens, we need to set ready to true again.
+            turnReady()
```
Mr3zee: Hmmm, even on an error?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeGrpcLibrary.kt
```diff
@@ -0,0 +1,42 @@
...
+
+    /** Acquire a runtime reference. Must be closed exactly once. */
+    fun acquire(): AutoCloseable {
+        refLock.withLock {
+            val prev = 0
+            refs++
+            if (prev == 0) grpc_init()
+        }
+        return object : AutoCloseable {
+            private var done = atomic(false)
```
Mr3zee: `var` -> `val`

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeGrpcLibrary.kt
```diff
@@ -0,0 +1,42 @@
...
+internal object GrpcRuntime {
+    private val refLock = reentrantLock()
+    private var refs = 0
+
+    /** Acquire a runtime reference. Must be closed exactly once. */
+    fun acquire(): AutoCloseable {
+        refLock.withLock {
+            val prev = 0
+            refs++
+            if (prev == 0) grpc_init()
```
Mr3zee: prev is always 0

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeManagedChannel.kt
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ * Insecure credentials.
+ */
+internal class GrpcInsecureCredentials() :
+    GrpcCredentials(grpc_insecure_credentials_create() ?: error("Failed to create credentials"))
+
+// default propagation mask for all calls.
+private const val GRPC_PROPAGATE_DEFAULTS = 0x0000FFFFu
```
Mr3zee: Is there a reference to this mask?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeManagedChannel.kt
```diff
@@ -0,0 +1,146 @@
...
+        ?: error("Failed to create channel")
+
+    @Suppress("unused")
+    private val rawCleaner = createCleaner(raw) {
+        grpc_channel_destroy(it)
+    }
+
+    override val platformApi: ManagedChannelPlatform = this
+
+    private var isShutdownInternal: Boolean = false
```
Mr3zee: Please make ThreadSafe

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeManagedChannel.kt
```diff
@@ -0,0 +1,146 @@
...
+    override val isShutdown: Boolean = isShutdownInternal
+    private var isTerminatedInternal = CompletableDeferred<Unit>()
```
Mr3zee: `var` -> `val`

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeManagedChannel.kt
```diff
@@ -0,0 +1,146 @@
...
+
+        val callJob = Job(callJobSupervisor)
+
+        val methodFullName = methodDescriptor.getFullMethodName()
+        // to construct a valid HTTP/2 path, we must prepend the name with a slash.
+        // the user does not do this to align it with the java implementation.
+        val methodNameSlice = "/$methodFullName".toGrpcSlice()
+        val rawCall = grpc_channel_create_call(
+            channel = raw, parent_call = null, propagation_mask = GRPC_PROPAGATE_DEFAULTS, completion_queue = cq.raw,
+            method = methodNameSlice, host = null, deadline = gpr_inf_future(GPR_CLOCK_REALTIME), reserved = null
```
Mr3zee: Please put parameters on the different lines

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/utils.kt
```diff
@@ -0,0 +1,167 @@
...
+}
+
+internal suspend fun withArena(block: suspend (Arena) -> Unit) =
+    Arena().let { arena ->
+        try {
+            block(arena)
+        } finally {
+            arena.clear()
+        }
+    }
```
Mr3zee: probably unused

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/utils.kt
```diff
@@ -0,0 +1,167 @@
...
+
+internal fun grpc_slice.toByteArray(): ByteArray = memScoped {
+    val out = ByteArray(len().toInt())
+    if (out.isEmpty()) return out
+
+    out.usePinned {
+        memcpy(it.addressOf(0), startPtr(), len().convert())
+    }
```
Mr3zee, line 43: Is there a way to avoid coping?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/ManagedChannel.native.kt
```diff
@@ -17,19 +20,42 @@ public actual abstract class ManagedChannelPlatform : GrpcChannel()
...
+
+    fun buildChannel(): NativeManagedChannel {
+        return NativeManagedChannel(
+            target,
+            credentials = credentials ?: error("No credentials set"),
+        )
+    }
+
+}
+
```
Mr3zee: is it possible you move the impl into `ManagedChannelBuilder` so that there is no need for the `error("Builder does not override usePlaintext()")` tricks?

protobuf/protobuf-core/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ kotlin {
                 api(projects.utils)
                 api(projects.protobuf.protobufInputStream)
                 api(projects.grpc.grpcCodec)
+                implementation("com.google.api.grpc:proto-google-common-protos:2.60.0")
```
Mr3zee: Please move into `libs.version.toml`

Mr3zee (changes requested):
Wow, great work! Thanks for the amount of comments and the big test set, it was pretty pleasant to read!

I have some comments and suggestions below, please address them before merging.

One big question that I have, is using C lib instead of C++ one - how hard it would be for use to support gRPC features as Auth, interceptors, etc? It is available is C API?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/CompletionQueue.kt
Jozott00: Yes, it is thread-safe, I will mention it in the comment.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/CompletionQueue.kt
Jozott00: The C lib only has the `shutdown()` function. The `forceShutdown` is defined by our implementation to prevent that any batches are submitted when we want to stop as soon as possible. I am not sure how to call it tho.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
Jozott00: Yeah, but callbacks from the completion queue are executed concurrently, so we have to be sure that we deliver the status only once.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
Jozott00: I guess it wouldn't be too hard. Not sure if it has a performance impact in the way we use the client call. However, I will create an issue to implement this.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
Jozott00: If there is a `closeInfo`, it is impossible that `inFlight` is incremented again. 
This is because when the `RECV_STATUS_ON_CLIENT` op finishes in the completion queue, all other OP batches must already be done in the completion queue and therefore their inFlight increment was also performed.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
Jozott00: In the `startRecvStatus` method. I added a little hint.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
Jozott00: All operations in a batch are handled as a unit, so the CQ callback is executed when all operations are done. Therefore, we only have to increment inFlight ones.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
Jozott00: Yep, if the call is canceled, the ongoing batches will execute this path.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeGrpcLibrary.kt
Jozott00: Ahh thanks!

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/utils.kt
Jozott00: Unfortunately not. However, this is currently only used for the statusDetails string which is only set if the call failed.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeManagedChannel.kt
Jozott00: It is defined as a macro in the grpc-c library and I thought I couldn't reference it with cinterop, but it actually works.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/ManagedChannel.native.kt
Jozott00: ~I think it's not possible. When I try, it says that there is no corresponding `expect` function.~

It is a bit odd to implement it in the abstract class, as I would have to cast `this` unchecked to `T`. Also, I can't override the constructor, so I would have to set the target nullable. The Java implementation also has an error thrown in the `ManagedChannel`.

protobuf/protobuf-core/build.gradle.kts
Jozott00: Ahh yes, forgot to remove it, we don't need it anymore. Thanks!

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
Jozott00: I limited the allowed number of requested messages.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/CompletionQueue.kt
```diff
@@ -0,0 +1,197 @@
...
+     * The returned [CallbackFuture] will be completed with `Unit` when the queue is shut down.
+     *
+     * @param force if `true`, the queue will reject all new batches with [BatchResult.CQShutdown].
+     *        Otherwise, the queue allows submitting new batches and shutdown only when there are no more
+     *        ongoing batches.
+     */
+    fun shutdown(force: Boolean = false): CallbackFuture<Unit> {
+        if (force) {
+            forceShutdown = true
+        }
```
Mr3zee: Not thread safe operation

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/CompletionQueue.kt
Mr3zee: Oh, ok, then it's fine

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
Mr3zee: Got it, thanks!

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/ManagedChannel.native.kt
```diff
@@ -17,19 +20,42 @@ public actual abstract class ManagedChannelPlatform : GrpcChannel()
...
     }
 }
 
+internal class NativeManagedChannelBuilder(
+    private val target: String,
+) : ManagedChannelBuilder<NativeManagedChannelBuilder>() {
+    private var credentials: GrpcCredentials? = null
+
+    override fun usePlaintext(): NativeManagedChannelBuilder {
+        check(credentials == null) { "Credentials already set" }
```
Mr3zee: I think it should be idempotent and not throw

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/ManagedChannel.native.kt
Mr3zee: aah, ok, nvm

Mr3zee (approved):
LGTM! 

Please look at the comments below
Also, please revisit all places where you do `check` and replace with `internalError` if needed, I saw at least one such place

PR #454, opened 2025-08-19
Protobuf conformance fixes

Mr3zee:
**Subsystem**
Protobuf

**Problem Description**
Some tests were failing due to non conformance

**Solution**
Fixed: 
- Repeated size calculation
- Default values
- Messages size calculation 
- Packed/unpacked decoding
- BytesUntilLimit check
- repeated fields

### DIFF
--- a/protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
+++ b/protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
@@ -20,7 +20,6 @@ import kotlinx.rpc.protoc.gen.core.model.Model
 import kotlinx.rpc.protoc.gen.core.model.OneOfDeclaration
 import kotlinx.rpc.protoc.gen.core.model.WireType
 import org.slf4j.Logger
-import kotlin.collections.map
 
 class ModelToProtobufKotlinCommonGenerator(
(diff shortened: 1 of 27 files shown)


protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -271,6 +308,8 @@ class ModelToProtobufKotlinCommonGenerator(
             }
         }
 
+        code("msg.checkRequiredFields()")
```
Jozott00: I placed the `checkRequiredFields()` call in the decode function to make it lazily checked (when the complete message decoding finished). I did this because if we check required fields directly after decoding a sub-message, a missing required field causes an error for that sub-message. However, I _think_ that in protobuf it is allowed to send message in chunks (i.e. a message can't be decoded at once, but in several iterations). In this case, the first message chunk might not contain the required field, but the second one does. However, when checking for required fields directly in the `decodeWith`, it will throw an error, as the required field is not present in the first chunk.

What was the reason for changing this? 
Sub-messages _should_ be recursively checked in `checkRequiredFields()`.

Mr3zee: Yes, you are right, thanks for pointing out!

PR #455, opened 2025-08-19
Update ktor-all-platforms-app sample to sync service creation

Mr3zee:
**Subsystem**
Sample

**Problem Description**
We made KtorRpcClient creation sync

**Solution**
Update sample to reflect that


samples/ktor-all-platforms-app/composeApp/src/commonMain/kotlin/App.kt
```diff
@@ -48,53 +46,49 @@ fun App() {
                     json()
                 }
             }
-        }.withService()
+        }.withService<UserService>()
```
dalewking: One of the uses of UserService in this statement is redundant

Mr3zee: Fixed, thanks!

samples/ktor-all-platforms-app/composeApp/src/commonMain/kotlin/App.kt
```diff
@@ -51,50 +49,46 @@ fun App() {
         }.withService()
```
dalewking: I think it is better practice to extract the `client.rpc { ... }` call into its own variable to emphasize that you don't need to make the rpc call every time:
```
    val rpcClient = remember {
        client.rpc {
            url {
                host = DEV_SERVER_HOST
                port = 8080
            }
        }
   }
   val service: UserService = remember { rpcClient.withService() }

Mr3zee: True, will do

e5l (approved):

PR #457, opened 2025-08-20
grpc: Fix service names in the generated descriptor

Jozott00:
**Subsystem**
gRPC

**Problem Description**
Currently the generated gRPC service name in the descriptor is just the simple names of the Kotlin interface. However, in gRPC a service name is structured as `<proto-package>.<name>`.
Additionally, the stub generator had no information about the `proto-package` set in the proto file, which is required to generate the correct service name. 

**Solution**
This PR adds the `protoPackage` argument to the `@Grpc` annotation. The service interface generator will set it if the `package` property is set in the proto file. 
This argument is then used by the service stub generator to obtain the proto package name.
To test this, the JavaPackageOptionTest was added, as well as an additional service interface in the compiler plugin tests.

Additionally, this PR moves the CoreClientTest into commonTest and provides service implementations that are required to run the tests. While it is not yet possible to automatically execute those tests, it got easier.

### DIFF
--- a/compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcDeclarationScanner.kt
+++ b/compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcDeclarationScanner.kt
@@ -4,6 +4,7 @@
 
 package kotlinx.rpc.codegen.extension
 
+import kotlinx.rpc.codegen.common.RpcClassId
 import kotlinx.rpc.codegen.common.RpcNames
 import org.jetbrains.kotlin.cli.common.messages.CompilerMessageSeverity
 import org.jetbrains.kotlin.cli.common.messages.MessageCollector
@@ -17,7 +18,6 @@ import org.jetbrains.kotlin.ir.expressions.IrExpression
 import org.jetbrains.kotlin.ir.util.dumpKotlinLike
 import org.jetbrains.kotlin.ir.util.getAnnotation
 import org.jetbrains.kotlin.ir.util.hasDefaultValue
-import org.jetbrains.kotlin.ir.util.kotlinFqName
 
 /**
  * This class scans user declared RPC service
@@ -30,7 +30,7 @@ internal object RpcDeclarationScanner {
     fun scanServiceDeclaration(service: IrClass, ctx: RpcIrContext, logger: MessageCollector): ServiceDeclaration {
         var stubClass: IrClass? = null
 
-        val grpcAnnotation = service.getAnnotation(ctx.grpcAnnotation.owner.kotlinFqName)
+        val grpcAnnotation = service.getAnnotation(RpcClassId.grpcAnnotation.asSingleFqName())
         val protoPackage = grpcAnnotation?.arguments?.getOrNull(0)?.asConstString() ?: ""
 
         val declarations = service.declarations.memoryOptimizedMap { declaration ->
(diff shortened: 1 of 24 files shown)


compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcDeclarationScanner.kt
```diff
@@ -25,6 +30,9 @@ internal object RpcDeclarationScanner {
     fun scanServiceDeclaration(service: IrClass, ctx: RpcIrContext, logger: MessageCollector): ServiceDeclaration {
         var stubClass: IrClass? = null
 
+        val grpcAnnotation = service.getAnnotation(ctx.grpcAnnotation.owner.kotlinFqName)
```
Mr3zee: This will fail on `ctx.grpcAnnotation` if the service is not gRPC with unresolved class error for `@Grpc`

Please add `isGrpc` check

compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcStubGenerator.kt
```diff
@@ -39,6 +35,8 @@ private object Stub {
 private object Descriptor {
     const val CALLABLE_MAP = "callableMap"
     const val CALLABLES = "callables"
+
+    // the kotlin class name (derived from java_package and name)
```
Mr3zee: FQ_NAME is not only for gRPC
`// the kotlin class fqName (not the same with protoService name for gRPC)`

compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcStubGenerator.kt
```diff
@@ -1105,7 +1104,7 @@ internal class RpcStubGenerator(
         }.apply {
             arguments {
                 values {
-                    +stringConst(declaration.simpleName)
+                    +stringConst(declaration.serviceFqName)
```
Mr3zee, line 1107: Please also change the comment above

tests/compiler-plugin-tests/src/testData/box/grpc.fir.ir.txt
```diff
@@ -71,7 +71,7 @@ FILE fqName:<root> fileName:/grpc.kt
         public open fun toString (): kotlin.String declared in kotlin.Any
   CLASS INTERFACE name:BoxService modality:ABSTRACT visibility:public superTypes:[kotlin.Any]
     annotations:
-      Grpc
+      Grpc(protoPackage = <null>)
```
Mr3zee, line 74: Shouldn't it be `""`?

Jozott00: Yeah, I have no clue why it says null and not "". I guess this is how unset arguments with default values are represented, so one can distinguish between set and not set.

grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/test/CoreClientTest.kt
```diff
@@ -236,13 +200,12 @@ class GrpcCoreTest {
...
-            override fun onClose(status: Status, trailers: GrpcTrailers) {
-                statusDeferred.complete(status)
-            }
-        }
+        val listener = createClientCallListener<HelloReply>(
+            onClose = { status, _ -> statusDeferred.complete(status) }
+        )
 
         channel.shutdown()
+        runBlocking { delay(100) }
```
Mr3zee: What is this for?

PR #458, opened 2025-08-21
grpc-native: Add support for iOS

Jozott00:
**Subsystem**
gRPC/Native/iOS

**Solution**
This PR adds iOS compilation of `protowire.a` and `kgrpc.a`.
iOS is now supported to the same level as macOS-Arm64.


cinterop-c/build_all_targets.sh
```diff
@@ -0,0 +1,52 @@
...
+OUTDIR="${2:?need output dir}"
+mkdir -p "$OUTDIR"
+
+# Compilation Config
+CONFIG=release
+
+# Platform labels
+MACOS=@build_bazel_apple_support//platforms:macos_arm64
+IOS_DEV=@build_bazel_apple_support//platforms:ios_arm64
+IOS_SIM=@build_bazel_apple_support//platforms:ios_sim_arm64
```
Mr3zee: What about iOS x64?

cinterop-c/MODULE.bazel
```diff
@@ -19,6 +22,6 @@ bazel_dep(
 # gRPC library
 bazel_dep(
     name = "grpc",
-    version = "1.73.1",
+    version = "1.74.1",
```
Mr3zee, line 25: Please sync this version with `libs.versions.toml`

gradle-conventions/src/main/kotlin/util/cinterop.kt
```diff
@@ -46,9 +45,9 @@ fun KotlinMultiplatformExtension.configureCLibCInterop(
     val buildCinteropCLib = project.tasks.register<Exec>("buildCinteropCLib") {
         group = "build"
         workingDir = cinteropCLib
-        commandLine("bash", "-c", "bazel build $bazelTask --config=release")
-        inputs.files(project.fileTree(cinteropCLib) { exclude("bazel-*/**") })
-        outputs.dir(cinteropCLib.resolve("bazel-bin"))
+        commandLine("bash", "-c", "${cinteropCLib}/build_all_targets.sh $bazelTask out")
+        inputs.files(project.fileTree(cinteropCLib) { exclude("bazel-*/**", "out/**") })
+        outputs.dir(cinteropCLib.resolve("out"))
```
Mr3zee: Is it possible to make this task per target? Like `buildCinteropCLib_macosArm64`, `buildCinteropCLib_iosArm64`, etc? So not to build all at once

Jozott00: Yeah, I can change that 👍

cinterop-c/build_all_targets.sh
Jozott00: Ah, good catch, I added it now.

PR #461, opened 2025-08-26
grpc-native: Pre-Built gRPC library to avoid long compilation times

Jozott00:
**Subsystem**
gRPC/Native, project setup

**Problem Description**
Before this, the gRPC core library was compiled from source when loading the gradle project.
With multiple targets this took at least 10 minutes.

**Solution**
We are now including gRPC as a pre-built static library, so it is not required to build it from source. 

We also tried to use a remote repository for this, but we faced several issues.


cinterop-c/tools/collect_headers.bzl
```diff
@@ -0,0 +1,91 @@
+HeaderInfo = provider(fields = ["headers_dir", "headers"])
```
Mr3zee, line 16: Can you add a general overview of what this file does in the top comment please?

Mr3zee (approved):
LGTM

PR #466, opened 2025-09-02
grpc-native: Add linux support

Jozott00:
**Subsystem**
gRPC/Native

**Problem Description**
Before this PR Linux (Arm64 and X64) were not supported.

**Solution**
This PR adds compilation support for Linux by providing a custom Bazel toolchain that uses the Konan (Kotlin/Native) toolchain under the hood. This guarantees that the produced static libraries are binary compatible with K/N.


cinterop-c/toolchain/cc_toolchain_config.bzl
```diff
@@ -0,0 +1,203 @@
...
+    ACTION_NAMES.cpp_compile,
+    ACTION_NAMES.assemble,
+    ACTION_NAMES.preprocess_assemble,
+    ACTION_NAMES.cpp_link_executable,
+    ACTION_NAMES.cpp_link_dynamic_library,
+    ACTION_NAMES.cpp_link_nodeps_dynamic_library,
+    ACTION_NAMES.cpp_link_static_library,
+]
+
+def _impl(ctx):
```
Mr3zee, line 27: Please add a guide to what is happening here

protobuf/protobuf-core/gradle.properties
```diff
@@ -4,8 +4,6 @@
 kotlinx.rpc.exclude.wasmWasi=true
 kotlinx.rpc.exclude.js=true
 kotlinx.rpc.exclude.wasmJs=true
-kotlinx.rpc.exclude.linuxArm64=true
-kotlinx.rpc.exclude.linuxX64=true
 kotlinx.rpc.exclude.macosX64=true
```
Mr3zee: Should we also exclude macosX64 here?

Mr3zee (approved):
Amazing, thanks!

protobuf/protobuf-core/gradle.properties
Jozott00: Ahh yes, thanks for noting!

PR #469, opened 2025-09-04
grpc-native: Add server implementation

Jozott00:
**Subsystem**
gRPC/Native

**Solution**
This PR adds a gRPC server implementation for K/N.
It uses the same C-Core callback API as the C++ callback server implementation. This way, we don't have to provide our own thread pool to request incoming calls from the server.

### DIFF
--- a/cinterop-c/include/kgrpc.h
+++ b/cinterop-c/include/kgrpc.h
@@ -57,6 +57,7 @@ bool kgrpc_iomgr_run_in_background();
  *
  * Wraps the internal C++ API `Server::SetRegisteredMethodAllocator()` to enable
  * callback-driven method dispatch via the Core C API.
+ * If the C++ API is exposed to the C API, this can be removed (https://github.com/grpc/grpc/issues/40632).
  *
  * When the gRPC Core needs to accept a new call for the specified method, it invokes:
  *   kgrpc_registered_call_allocation alloc = allocator();
--- a/grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/test/CoreClientTest.kt
+++ b/grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/test/CoreClientTest.kt
@@ -6,6 +6,7 @@ package kotlinx.rpc.grpc.test
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.test.runTest
 import kotlinx.coroutines.withTimeout
 import kotlinx.rpc.grpc.GrpcServer
 import kotlinx.rpc.grpc.GrpcTrailers
@@ -270,21 +271,20 @@ class GreeterServiceImpl : GreeterService {
      * Run this on JVM before executing tests.
      */
     @Test
-    fun runServer() {
-        runBlocking {
-            val server = GrpcServer(
-                port = PORT,
-                builder = { registerService<GreeterService> { GreeterServiceImpl() } }
-            )
+    fun runServer() = runTest {
+        val server = GrpcServer(
+            port = PORT,
+            builder = { registerService<GreeterService> { GreeterServiceImpl() } }
+        )
 
-            try {
-                server.start()
-                println("Server started")
-                server.awaitTermination()
-            } finally {
-                server.shutdown()
-                server.awaitTermination()
-            }
+        try {
+            server.start()
+            println("Server started")
+            server.awaitTermination()
+        } finally {
+            server.shutdown()
+            server.awaitTermination()
+
         }
     }
 
--- a/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServer.kt
+++ b/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServer.kt
@@ -70,7 +70,8 @@ internal class NativeServer(
 
     private val cq = CompletionQueue()
 
-    val raw: CPointer<grpc_server> = grpc_server_create(null, null)!!
+    val raw: CPointer<grpc_server> = grpc_server_create(null, null)
+        ?: error("Failed to create server")
 
     // holds all stable references to MethodAllocationCtx objects.
     // the stable references must eventually be disposed.
--- a/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServerCall.kt
+++ b/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServerCall.kt
@@ -59,12 +59,12 @@ internal class NativeServerCall<Request, Response>(
         grpc_call_unref(it)
     }
 
-    private var listener = DeferredCallListener<Request>()
+    private val listener = DeferredCallListener<Request>()
     private var methodDescriptor: MethodDescriptor<Request, Response>? = null
-    private var callbackMutex = ReentrantLock()
+    private val callbackMutex = ReentrantLock()
     private var initialized = false
     private var cancelled = false
-    private var finalized = atomic(false)
+    private val finalized = atomic(false)
 
     // Tracks whether at least one request message has been received on this call.
     private var receivedFirstMessage = false
@@ -190,7 +190,7 @@ internal class NativeServerCall<Request, Response>(
 
     override fun request(numMessages: Int) {
         check(initialized) { internalError("Call not initialized") }
-        // TODO: Remove the num constraint
+        // TODO: Remove the num constraint (KRPC-213)
         require(numMessages == 1) { internalError("numMessages must be 1") }
         val methodDescriptor = checkNotNull(methodDescriptor) { internalError("Method descriptor not set") }
 
@@ -320,7 +320,7 @@ private class DeferredCallListener<T> : ServerCall.Listener<T>() {
     @Volatile
     private var delegate: ServerCall.Listener<T>? = null
     private val mutex = ReentrantLock()
-    private val q = ArrayDeque<(ServerCall.Listener<T>) -> Unit>()
+    private val queue = ArrayDeque<(ServerCall.Listener<T>) -> Unit>()
 
     fun setDelegate(d: ServerCall.Listener<T>) {
         mutex.withLock {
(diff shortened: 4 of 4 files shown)


cinterop-c/include/kgrpc.h
```diff
@@ -24,13 +24,73 @@ typedef struct {
 } kgrpc_cb_tag;
 
 
- /*
- * Call to grpc_iomgr_run_in_background(), which is not exposed as extern "C" and therefore must be wrapped.
+// This is a duplicate of the RegisteredCallAllocation, which is defined in
+// https://github.com/grpc/grpc/blob/893bdadd56dbb75fb156175afdaa2b0d47e1c15b/src/core/server/server.h#L150-L157.
+// This is required, as RegisteredCallAllocation is not part of the exposed C API.
```
Mr3zee, line 29: What are the chances that breaks because they changed the internal structure?
Are there any more safe workarounds?

cinterop-c/include/kgrpc.h
```diff
@@ -24,13 +24,73 @@ typedef struct {
...
+/*
+* Call to grpc_iomgr_run_in_background(), which is not exposed as extern "C" and therefore must be wrapped.
+*/
+bool kgrpc_iomgr_run_in_background();
+
+/**
+ * Registers a C-style allocator callback for accepting gRPC calls to a specific method.
+ *
+ * Wraps the internal C++ API `Server::SetRegisteredMethodAllocator()` to enable
+ * callback-driven method dispatch via the Core C API.
```
Mr3zee, line 59: What are the implications of the fact that we use internal API here?

grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/test/CoreClientTest.kt
```diff
@@ -259,19 +270,21 @@ class GreeterServiceImpl : GreeterService {
      * Run this on JVM before executing tests.
      */
     @Test
-    fun runServer() = runTest(timeout = Duration.INFINITE) {
-        val server = GrpcServer(
-            port = PORT,
-            builder = { registerService<GreeterService> { GreeterServiceImpl() } }
-        )
+    fun runServer() {
+        runBlocking {
```
Mr3zee: Better the use `runTest` if it is a test, or if not - remote the `@Test` annotation

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServer.kt
```diff
@@ -0,0 +1,243 @@
...
+    val fallbackRegistry: HandlerRegistry,
+) : Server {
+
+    // a reference to make sure the grpc_init() was called. (it is released after shutdown)
+    @Suppress("unused")
+    private val rt = GrpcRuntime.acquire()
+
+    private val cq = CompletionQueue()
+
+    val raw: CPointer<grpc_server> = grpc_server_create(null, null)!!
```
Mr3zee: We don't really use `!!` in codebases as it is error prone.
If it can never be null, better to use `?: error("unreachable")` or even add a full explanation in error message why it is unreachable

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServer.kt
```diff
@@ -0,0 +1,243 @@
...
+
+    // holds all stable references to MethodAllocationCtx objects.
+    // the stable references must eventually be disposed.
+    private val callAllocationCtxs = mutableSetOf<StableRef<CallAllocationCtx>>()
+
+    init {
+        grpc_server_register_completion_queue(raw, cq.raw, null)
+        grpc_server_add_http2_port(raw, "0.0.0.0:$port", credentials.raw)
```
Mr3zee, line 82: It is always `0.0.0.0`?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServerCall.kt
```diff
@@ -0,0 +1,357 @@
...
+    ) : this(raw, cq) {
+        setMethodDescriptor(methodDescriptor)
+    }
+
+    @Suppress("unused")
+    private val rawCleaner = createCleaner(raw) {
+        grpc_call_unref(it)
+    }
+
+    private var listener = DeferredCallListener<Request>()
```
Mr3zee: probably not `var`

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServerCall.kt
```diff
@@ -0,0 +1,357 @@
...
+    private var methodDescriptor: MethodDescriptor<Request, Response>? = null
+    private var callbackMutex = ReentrantLock()
```
Mr3zee: Probably not `var`

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServerCall.kt
```diff
@@ -0,0 +1,357 @@
...
+    private var initialized = false
+    private var cancelled = false
+    private var finalized = atomic(false)
```
Mr3zee: `var`

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServerCall.kt
```diff
@@ -0,0 +1,357 @@
...
+                    "Batch could not be submitted: ${result.error}"
+                )
+            }
+        }
+    }
+
+    override fun request(numMessages: Int) {
+        check(initialized) { internalError("Call not initialized") }
+        // TODO: Remove the num constraint
+        require(numMessages == 1) { internalError("numMessages must be 1") }
```
Mr3zee: Please remove the constraint or file an issue

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServerCall.kt
```diff
@@ -0,0 +1,357 @@
...
+            if (delegate != null) return
+            delegate = d
+        }
+        // drain the queue
+        q.forEach { it(d) }
+        q.clear()
+    }
+
+    private inline fun deliver(crossinline f: (ServerCall.Listener<T>) -> Unit) {
+        val d = delegate
```
Mr3zee: please name `currentDelegate`

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServerCall.kt
```diff
@@ -0,0 +1,357 @@
...
+        mutex.withLock {
+            if (delegate != null) return
+            delegate = d
+        }
+        // drain the queue
+        q.forEach { it(d) }
+        q.clear()
+    }
+
+    private inline fun deliver(crossinline f: (ServerCall.Listener<T>) -> Unit) {
```
Mr3zee: Please give a meaning full name to `f`

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServerCall.kt
```diff
@@ -0,0 +1,357 @@
...
+ * This class extends `ServerCall.Listener`, allowing it to serve as a listener for server calls.
+ * Initially, incoming method calls (e.g., `onMessage`, `onHalfClose`, etc.) are queued until a delegate
+ * is assigned through the `setDelegate` method. Once the delegate is set, queued methods are delivered
+ * in order and all future method calls are forwarded directly to the delegate.
+ */
+private class DeferredCallListener<T> : ServerCall.Listener<T>() {
+    @Volatile
+    private var delegate: ServerCall.Listener<T>? = null
+    private val mutex = ReentrantLock()
+    private val q = ArrayDeque<(ServerCall.Listener<T>) -> Unit>()
```
Mr3zee: Please name `queue`

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/serverCallTags.kt
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    override fun run(ok: Boolean) {
+        try {
+            if (!ok) {
+                return
+            }
+
+            // TODO: check authority
+            // val host = rawDetails.host.toByteArray().decodeToString()
```
Mr3zee, line 124: This is ssl stuff, right?

Mr3zee (changes requested):
Lightning fast! 
I pressed request changes as I want to discuss the questions below

cinterop-c/include/kgrpc.h
Jozott00: I think it _should_ be pretty stable (there was 1 little change 4 years ago). 
Unfortunately, the other available (C) APIs don’t play very well with the callback approach. It would require us to maintain our own thread pool with up/down scaling and the overall logic would increase a lot. 

This is the API the C++ callback server API is using.

cinterop-c/include/kgrpc.h
Jozott00: The API is not protected from breaking changes. 
However, the `Server::SetRegisteredMethodAllocator()` function did not change since its introduction in 2020 (when they introduced the Callback API).

grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/test/CoreClientTest.kt
Jozott00: Ah thanks! A left over from manual tests

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServer.kt
Jozott00: In java gRPC the builder when only the `port` is given looks like this:
```java
    public static NettyServerBuilder forPort(int port) {
        return forAddress(new InetSocketAddress(port));
    }
```

This will either be `0.0.0.0` or `::0` depending on whether IPv6 is supported or not. 
I don't know yet how to detect if IPv6 is supported, so I currently just use the IPv4 variant. To support more than the local address we would have to add some common builder API.

I will add a TODO and create an issue for supporting IPv6 detection.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeServerCall.kt
Jozott00: Good catch, thanks!

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/serverCallTags.kt
Jozott00: Yes, this will be implemented in the next PR.

PR #471, opened 2025-09-09
grpc: Add TLS credentials support

Jozott00:
**Subsystem**
gRPC

**Problem Description**
Previously all connections were plaintext and thus not encrypted.

**Solution**
This PR adds TLS encryption support for both client and server on JVM and Native. 
New TLS credentials can be created using `TlsClientCredentials` and `TlsServerCredentials`.

### DIFF
--- a/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeManagedChannel.kt
+++ b/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeManagedChannel.kt
@@ -64,7 +64,10 @@ internal class NativeManagedChannel(
 
     internal val raw: CPointer<grpc_channel> = memScoped {
         val args = authority?.let {
-            var authorityOverride = alloc<grpc_arg> {
+            // the C Core API doesn't have a way to override the authority (used for TLS SNI) as it
+            // is available in the Java gRPC implementation.
+            // instead, it can be done by setting the "grpc.ssl_target_name_override" argument.
+            val authorityOverride = alloc<grpc_arg> {
                 type = grpc_arg_type.GRPC_ARG_STRING
                 key = "grpc.ssl_target_name_override".cstr.ptr
                 value.string = authority.cstr.ptr


grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcServer.kt
```diff
@@ -49,6 +48,7 @@ private typealias ResponseServer = Any
  */
 public class GrpcServer internal constructor(
     override val port: Int = 8080,
+    credentials: ServerCredentials? = null,
```
Jozott00, line 51: @Mr3zee The `GrpcServer` constructor now takes an optional `credentials` argument because the gRPC `ServerBuilder` also only takes the credentials at construction time. It is not possible to set them using a builder method.

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcClient.kt
```diff
@@ -124,10 +131,11 @@ public class GrpcClient internal constructor(
 public fun GrpcClient(
     hostname: String,
     port: Int,
+    credentials: ClientCredentials? = null,
```
Jozott00, line 134: @Mr3zee The `GrpcClient` constructor now takes an optional credentials argument because the gRPC `ManagedChannelBuilder` also only takes the credentials at construction time. It is not possible to set them using a builder method.

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeManagedChannel.kt
```diff
@@ -73,8 +62,22 @@ internal class NativeManagedChannel(
     // the channel's completion queue, handling all request operations
     private val cq = CompletionQueue()
 
-    internal val raw: CPointer<grpc_channel> = grpc_channel_create(target, credentials.raw, null)
-        ?: error("Failed to create channel")
+    internal val raw: CPointer<grpc_channel> = memScoped {
+        val args = authority?.let {
+            var authorityOverride = alloc<grpc_arg> {
+                type = grpc_arg_type.GRPC_ARG_STRING
+                key = "grpc.ssl_target_name_override".cstr.ptr
```
Mr3zee, line 72: Can you make it a const please and write a description for it?

Mr3zee (commented):
Blazing fast, thanks!

PR #472, opened 2025-09-10
protobuf: Add support for deprecated groups

Jozott00:
**Subsystem**
Protobuf

**Problem Description**
Groups were not supported or implemented, and they even caused a crash when the wire contained a GROUP tag. 

**Solution**
The PR implements group support, which resulted in 10 more tests passing the conformance tests, specifically proto3 tests that validate the correct parsing error behavior for invalid `END_GROUP` tests.

However, functionality is also implemented for proto2 and editions that actually use groups in their proto file.

```
=== Conformance Test Results ===
Total baseline tests:   5545
[+] Passed tests:       1392
[-] Failed tests:       4153
```


Mr3zee:
@Jozott00 looks like auto-lint on commit was applied to some generated classes

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -729,14 +771,21 @@ class ModelToProtobufKotlinCommonGenerator(
...
                 // packed fields also have the tag + len
                 field.dec.isPacked -> code("__result += $valueSize.let { $tagSize + ${int32SizeCall("it")} + it }")
                 else -> code("__result += $valueSize")
             }
 
-            is FieldType.Message,
+            is FieldType.Message -> if (fieldType.dec.value.isGroup) {
+                val groupTagSize = tagSizeCall(field.number, WireType.START_GROUP)
+                // the group message size is the size of the message plus the start and end group tags
+                code("__result += $valueSize.let { (2 * $groupTagSize) + it }")
```
Mr3zee, line 784: I know it's a compile time constant anyway, but shouldn't it be `${2 * groupTagSize}`?😅

Mr3zee: Oh, wait, no
Then the question is - should we replace `tagSizeCall` as code gen to just a constant?

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -291,25 +298,49 @@ class ModelToProtobufKotlinCommonGenerator(
...
+            contextReceiver = "${declaration.internalClassFullName()}.Companion",
+            returnType = "Unit",
+        ) {
+            whileBlock("true") {
+                if (declaration.isGroup) {
+                    code("val tag = decoder.readTag() ?: throw ProtobufDecodingException(\"Missing END_GROUP tag for field: \${startGroup.fieldNr}.\")")
+                    ifBranch(condition = "tag.wireType == $PB_PKG.WireType.END_GROUP", ifBlock = {
+                        ifBranch(condition = "tag.fieldNr != startGroup.fieldNr", ifBlock = {
+                            code("throw ProtobufDecodingException(\"Wrong END_GROUP tag. Expected \${startGroup.fieldNr}, got \${tag.fieldNr}.\")")
+                        })
```
Mr3zee, line 320: I haven't seen this piece of code anywhere generated

Mr3zee (approved):
⚡️⚡️⚡️

Do we enable proto2 tests and editions then?

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: Yeah you are right, we should calculate the tag size at generation time (in all cases)!

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: Yeah that is because the proto2 conformance tests are not yet activated in this PR and proto3 doesn't allow groups.

PR #478, opened 2025-09-16
Kotlin 2.2.20

Mr3zee:
**Subsystem**
Gradle

**Problem Description**
Update to 2.2.20

**Solution**
In 2.2.0 we switched Yarn -> Npm because of KT-78504.
Now the issue is resolved, we switch back, because in 2.3.0 (Kotlin Master builds) Npm started misbehaving


DestBro:
There is a patch version of kotlin: 2.2.21 - can an artifact for KRPC be done as well @Mr3zee ?

PR #481, opened 2025-09-17
Fix flaky tests

Mr3zee:
**Subsystem**
kRPC

**Problem Description**
Some tests are flaky on js/wasmJs platforms on CI because they are slow

**Solution**
Decrease iterations, increase timeouts


e5l (approved):
Please check CI, it looks like some tests still flaky

PR #484, opened 2025-09-18
grpc: Add interceptor API

Jozott00:
**Subsystem**
gRPC

**Problem Description**
A main feature of all gRPC libraries is their interceptor API.
We want a Kotlin idiomatic simple-to-use interceptor API (for client and server) that allows users to implement things like authentication, logging or tracing.

**Solution**
This PR adds a convenient Kotlin idiomatic interceptor API.


grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/suspendClientCalls.kt
```diff
@@ -261,3 +201,142 @@ internal class Ready(private val isReallyReady: () -> Boolean) {
...
+        onHeadersFuture.onComplete { block(it) }
+    }
+
+    override fun onClose(block: (Status, GrpcMetadata) -> Unit) {
+        onCloseFuture.onComplete { block(it.first, it.second) }
+    }
+
+    override fun cancel(message: String, cause: Throwable?): Nothing {
+        throw StatusException(Status(StatusCode.CANCELLED, message, cause))
+    }
```
Mr3zee, line 229: This is a bit unexpected for how usually `cancel` works in Kotlin. Is there any other way to cancel, without throwing an exception here?

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/suspendServerCalls.kt
```diff
@@ -144,19 +157,29 @@ private fun <Request, Response> CoroutineScope.serverCallListenerImpl(
...
+    val serverCallScope = ServerCallScopeImpl(
+        method = descriptor,
+        interceptors = interceptors,
+        implementation = implementation,
+        requestHeaders = requestHeaders,
+        serverCall = handler,
+        grpcContext = context.grpcContext,
+    )
+
+    val rpcJob = launch() {
```
Mr3zee: unused `()`

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/suspendServerCalls.kt
```diff
@@ -242,4 +266,41 @@ private class ServerCallListenerState {
...
+    val onCloseFuture = CallbackFuture<Pair<Status, GrpcMetadata>>()
+
+    override fun onClose(block: (Status, GrpcMetadata) -> Unit) {
+        onCloseFuture.onComplete { block(it.first, it.second) }
+    }
+
+    override fun close(status: Status, trailers: GrpcMetadata): Nothing {
+        // this will be cached by the rpcImpl() runCatching{} and turns it into a close()
+        throw StatusException(status, trailers)
+    }
```
Mr3zee, line 295: Same as previous. Another idea: maybe a different name?

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcClient.kt
```diff
@@ -131,23 +128,57 @@ public class GrpcClient internal constructor(
...
+        config.overrideAuthority?.let { overrideAuthority(it) }
+    }.buildChannel()
+    return GrpcClient(channel, config.messageCodecResolver, config.interceptors)
 }
+
+public class GrpcClientConfiguration internal constructor() {
+    internal var messageCodecResolver: MessageCodecResolver = EmptyMessageCodecResolver
+    internal var credentials: ClientCredentials? = null
+    internal var overrideAuthority: String? = null
+    internal val interceptors: MutableList<ClientInterceptor> = mutableListOf()
```
Mr3zee, line 192: Couple of comment os how usually DSLs are written in Kotlin. This design is good too, but let's make it more idiomatic:
- expose `messageCodecResolver` as a `var`, with no function
- expose `credentials` and do something like `credentials = plaintText() /* shortcut */; credentials = <object of type ClientCredentials>`
- expose `overrideAuthority` as a `var`
- `intercept()` is good imo

Same for `GrpcServerConfiguration`

Also please add kDocs to public methods/properties

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/ManagedChannel.kt
```diff
@@ -68,6 +68,8 @@ public interface ManagedChannel {
  * Builder class for [ManagedChannel].
  */
 public expect abstract class ManagedChannelBuilder<T : ManagedChannelBuilder<T>> {
+
+    // TODO: Not used anymore
     public fun usePlaintext(): T
```
Mr3zee: Should we remove it then?

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/ServerInterceptor.kt
```diff
@@ -0,0 +1,126 @@
...
+     */
+    public val responseTrailers: GrpcMetadata
+
+    /**
+     * The [GrpcContext] associated with this call.
+     *
+     * It can be used by the interceptor to provide call-scoped information about
+     * the current call, such as the identity of the caller or the current authentication state.
+     */
+    public val grpcContext: GrpcContext
```
Mr3zee: Probably just `context` would suffice

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/ServerInterceptor.kt
```diff
@@ -0,0 +1,126 @@
...
+     *   [proceed]. You may also transform the resulting response [Flow] before returning it to the framework.
+     * - The interceptor must ensure to provide and return a valid number of messages, depending on the method type.
+     * - The interceptor must not throw an exception. Use [close] to terminate the call with an error.
+     */
+    public fun proceed(request: Flow<Request>): Flow<Response>
+
+    /**
+     * Convenience for flow builders: proceeds with [request] and emits the resulting response elements into this
+     * [FlowCollector]. Useful inside `flow {}` blocks within interceptors.
+     */
```
Mr3zee, line 103: Let's add an example

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/ServerInterceptor.kt
```diff
@@ -0,0 +1,126 @@
...
+    public suspend fun FlowCollector<Response>.proceedFlow(request: Flow<Request>) {
```
Mr3zee: Let's call it `proceedUnmodified`

grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/test/BaseGrpcServiceTest.kt
```diff
@@ -30,16 +30,19 @@ abstract class BaseGrpcServiceTest {
...
         val server = GrpcServer(
             port = PORT,
-            messageCodecResolver = resolver,
             parentContext = coroutineContext,
+            configure = {
+                useMessageCodecResolver(resolver)
+            },
             builder = {
                 registerService(kClass) { impl }
             }
```
Mr3zee: Maybe we can unify `configure` and `builder`?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -408,6 +425,18 @@ internal class NativeClientCall<Request, Response>(
...
+     * Safely executes the provided block of user code, catching any thrown exceptions or errors.
+     * If an exception is caught, it cancels the operation with the specified message and cause.
+     */
+    private fun safeUserCode(cancelMsg: String, block: () -> Unit) {
+        try {
+            block()
+        } catch (e: Throwable) {
+            cancel(cancelMsg, e)
+        }
+    }
```
Mr3zee, line 440: Do we have the same behaviour on jvm?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -133,18 +133,19 @@ internal class NativeClientCall<Request, Response>(
...
      * Sets the [closeInfo] and calls [tryToCloseCall].
      * This is called as soon as the RECV_STATUS_ON_CLIENT batch (started with [startRecvStatus]) finished.
      */
-    private fun markClosePending(status: Status, trailers: GrpcTrailers) {
-        if (closeInfo.compareAndSet(null, Pair(status, trailers))) {
-            tryToCloseCall()
-        }
+    private fun markClosePending(status: Status, trailers: GrpcMetadata) {
+        closeInfo.compareAndSet(null, Pair(status, trailers))
+        tryToCloseCall()
```
Mr3zee, line 148: Is it right that this can be called multiple times now?

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
```diff
@@ -353,20 +363,26 @@ internal class NativeClientCall<Request, Response>(
 
     override fun cancel(message: String?, cause: Throwable?) {
         cancelled = true
-        val message = if (cause != null) "$message: ${cause.message}" else message
-        cancelInternal(grpc_status_code.GRPC_STATUS_CANCELLED, message ?: "Call cancelled")
+        val status = Status(StatusCode.CANCELLED, message ?: "Call cancelled", cause)
+        // user side cancellation must always win over any other status (even if the call is already completed).
+        // this will also preserve the cancellation cause, which cannot be passed to the grpc-core.
+        closeInfo.value = Pair(status, GrpcMetadata())
```
Mr3zee, line 370: Should we do something like `already cancelled` behavior? If a user cancels twice?

Mr3zee: Also, here you may have a race, if a user cancel and a system cancel happen together, but user sets the status first, then it will be overriden by the system

Mr3zee: Btw, do we have cancellation tests? This is different topic, but do we check what happens when a user cancels the coroutine in which the call is happening?

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/suspendClientCalls.kt
Jozott00: We can also cancel the message using the handler, however, in that case, `cancel` would not return `Nothing`. The outcome is the same except that the user has more logic in the interceptor as the control flow is not aborted when calling `cancel`.

Mr3zee: > the control flow is not aborted

I think that is fine, as per coroutines:
image

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
Jozott00: I think there is no "right" handling of double cancellation on the user side. Currently the last call wins. 

System cancellation doesn't call this method but `cancelInternal`, on the suspendClientCalls side, so there shouldn't be such a race.

Yeah, I have quite some cancellation and error throwing tests, for both client and server side interceptors. (`ServerInterceptorTest` and `ClientInterceptorTest`).

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/internal/NativeClientCall.kt
Jozott00: Yes, I tested it

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcServer.kt
```diff
@@ -187,17 +198,147 @@ public class GrpcServer internal constructor(
...
+ *
+ * This class provides an API to configure various server parameters, such as message codecs,
+ * security credentials, server-side interceptors, and service registration.
+ */
+public class GrpcServerConfiguration internal constructor() {
+
+    internal val interceptors: MutableList<ServerInterceptor> = mutableListOf()
+    internal var serviceBuilder: RpcServer.() -> Unit = { }
+
+
```
Mr3zee, line 254: Please leave only one empty line in between declarations, same in some other places

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcServer.kt
```diff
@@ -187,17 +198,147 @@ public class GrpcServer internal constructor(
...
+    /**
+     * Configures the gRPC server to register services.
+     *
+     * This method allows defining a block of logic to configure an [RpcServer] instance,
+     * where multiple services can be registered:
+     * ```kt
+     * GrpcServer(port) {
+     *     services {
+     *         registerService(MyService())
+     *         registerService(MyOtherService())
```
Mr3zee: This is a bit wrong example, should be:
```kotlin
registerService<MyService> { MyServiceImple() } 
```

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcServer.kt
```diff
@@ -187,17 +198,147 @@ public class GrpcServer internal constructor(
...
+ * To start the server, call the [GrpcServer.start] method.
+ * To clean up resources, call the [GrpcServer.shutdown] or [GrpcServer.shutdownNow] methods.
+ *
+ * ```kt
+ * GrpcServer(port) {
+ *     credentials = tls(myCertChain, myPrivateKey)
+ *     services {
+ *         registerService(MyService())
+ *         registerService(MyOtherService())
+ *     }
```
Mr3zee: Same here

grpc/grpc-core/src/commonTest/kotlin/kotlinx/rpc/grpc/test/RawClientTest.kt
```diff
@@ -54,8 +59,8 @@ class RawClientTest {
     fun clientStreamingEchoTest() = runTest(
         methodName = "ClientStreamingEcho",
         type = MethodType.CLIENT_STREAMING,
-    ) { channel, descriptor ->
-        val response = clientStreamingRpc(channel, descriptor, flow {
+    ) { client, descriptor ->
+        val response = client.clientStreamingRpc(descriptor, flow {
             repeat(5) {
                 delay(100)
```
Mr3zee: Please remove `delay` from tests, replace with completable deferred

Mr3zee (approved):
Great, thank you!

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcServer.kt
Jozott00: Upsi

Mr3zee:
@morki in some sort, yes, though later, we only maintain kRPC now, until we finish with gRPC (which is not much else to do, but still will take a couple of months)

PR #487, opened 2025-09-18
Grpc.Method Annotation

Mr3zee:
**Subsystem**
gRPC, Compiler Plugin

**Problem Description**
Java and other impls support `safe`, `idempotent` and `sampledToLocalTracing` parameters for methods

**Solution**
Support them and also method name + add custom chekers


compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/ServiceDeclaration.kt
```diff
@@ -33,10 +37,22 @@ class ServiceDeclaration(
     }
 
     class Method(
+        val ctx: RpcIrContext,
```
Jozott00: I think the new `ctx` property is unused.

Jozott00 (approved):
LGTM!

compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/ServiceDeclaration.kt
Mr3zee: True, leftovers

PR #490, opened 2025-09-29
PB: Added hasCode, equals and toString support

Mr3zee:
**Subsystem**
Protoc-gen

**Problem Description**
Out generated messages don't have any hashCode, equals and toString

**Solution**
Make pb-specific implementations of these functions

### DIFF
--- a/protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
+++ b/protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
@@ -285,11 +285,14 @@ class ModelToProtobufKotlinCommonGenerator(
             addLine("if (other == null || this::class != other::class) return false")
             addLine("other as ${declaration.internalClassName()}")
             addLine("other.checkRequiredFields()")
+            if (declaration.presenceMaskSize != 0) {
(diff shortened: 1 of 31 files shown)


protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -191,6 +196,185 @@ class ModelToProtobufKotlinCommonGenerator(
...
+            addLine("checkRequiredFields()")
+            addLine("if (this === other) return true")
+            addLine("if (other == null || this::class != other::class) return false")
+            addLine("other as ${declaration.internalClassName()}")
+            addLine("other.checkRequiredFields()")
+            if (fields.isNotEmpty()) {
+                fields.forEach { field ->
+                    if (field.presenceIdx != null) {
+                        fieldEqualsCheck(
+                            presenceCheck = "presenceMask[${field.presenceIdx}] != other.presenceMask[${field.presenceIdx}] || presenceMask[${field.presenceIdx}] && ",
```
Jozott00: It might be simpler and more performant to add a `presenceMask == other.presenceMask` before the fields check, instead of comparing the presence for every field.

Mr3zee: Great thinking! Will do that

PR #504, opened 2025-10-08
grpc: Splitting core module into client and server

Jozott00:
**Subsystem**
gRPC

**Problem Description**
Currently, almost all gRPC related code is in the `:grpc:grpc-core` module.

**Solution**
This PR splits the code into `grpc-core, grpc-client` and `grpc-server` modules.


grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/internal/suspendUtils.kt
```diff
@@ -0,0 +1,67 @@
...
+import kotlinx.coroutines.channels.onFailure
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.single
+import kotlinx.rpc.grpc.Status
+import kotlinx.rpc.grpc.StatusCode
+import kotlinx.rpc.grpc.StatusException
+import kotlinx.rpc.internal.utils.InternalRpcApi
+
+// todo really needed?
```
Mr3zee: probably this todo is obsolete

PR #517, opened 2025-10-21
grpc: Add Call Metadata (Headers and Trailers)

Jozott00:
**Subsystem**
gRPC

**Problem Description**
Before this, we had no support for passing gRPC metadata between the client and the server.

**Solution**
This PR solves this on JVM and Native by providing a custom interface (`GrpcMetadata`).

### DIFF
--- a/grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/GrpcMetadata.jvm.kt
+++ b/grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/GrpcMetadata.jvm.kt
@@ -13,7 +13,7 @@ import kotlinx.rpc.internal.utils.InternalRpcApi
 @InternalRpcApi
 public actual typealias GrpcMetadata = io.grpc.Metadata
 
-public actual class GrpcMetadataKey<T> internal actual constructor(
+public actual class GrpcMetadataKey<T> public actual constructor(
     private val name: String,
     private val codec: MessageCodec<T>,
 ) {
@@ -54,7 +54,7 @@ public actual operator fun GrpcMetadata.get(key: String): String? {
     return get(Metadata.Key.of(key, Metadata.ASCII_STRING_MARSHALLER))
 }
 
-public actual fun <T> GrpcMetadata.get(key: GrpcMetadataKey<T>): T? {
+public actual operator fun <T> GrpcMetadata.get(key: GrpcMetadataKey<T>): T? {
     return get(key.toAsciiKey())
 }
 
--- a/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/GrpcMetadata.native.kt
+++ b/grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/GrpcMetadata.native.kt
@@ -33,12 +33,8 @@ import libkgrpc.grpc_slice_unref
 import libkgrpc.kgrpc_metadata_array_append
 import kotlin.experimental.ExperimentalNativeApi
 
-public actual class GrpcMetadataKey<T> actual constructor(public var name: String, public val codec: MessageCodec<T>) {
-
-    init {
-        name = name.lowercase()
-    }
-
+public actual class GrpcMetadataKey<T> actual constructor(name: String, public val codec: MessageCodec<T>) {
+    public val name: String = name.lowercase()
     internal val isBinary get() = name.endsWith("-bin")
 
     internal fun encode(value: T): ByteArray = codec.encode(value).buffer.readByteArray()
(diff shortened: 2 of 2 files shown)


grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcMetadata.kt
```diff
@@ -4,7 +4,211 @@
...
-    public fun merge(trailers: GrpcMetadata)
-}
+public expect class GrpcMetadata constructor()
+
+/**
+ * Returns the last metadata entry added with the given [key], or `null` if there are no entries.
+ *
+ * @param key the name of the metadata entry to retrieve (case-insensitive). Must not end with `-bin`.
+ * @return the last value associated with the key, or `null` if no values exist
+ * @throws IllegalArgumentException if the key ends with `-bin`
```
Mr3zee: Probably also when the key doesn't conform to this:
```
Keys must contain only the following ASCII characters:
 * - Digits: `0-9`
 * - Lowercase letters: `a-z` (uppercase letters are normalized to lowercase)
 * - Special characters: `-`, `_`, `.`
```

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcMetadata.kt
```diff
@@ -4,7 +4,211 @@
...
+ * Non-ASCII characters in values are replaced with `?` during encoding.
+ *
+ * ## Thread Safety
+ *
+ * This class is not thread-safe. Modifications made by one thread may not be visible to another
+ * thread concurrently reading the metadata.
+ *
+ * ## Example usage
+ * ```kotlin
+ * val metadata = GrpcMetadata().apply {
```
Mr3zee, line 41: Let's also add `buildGrpcMetadata(body)` method, which does the same as `GrpcMetadata().apply { body() }`

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/GrpcMetadata.native.kt
```diff
@@ -2,9 +2,211 @@
...
+import libkgrpc.grpc_metadata_array
+import libkgrpc.grpc_metadata_array_init
+import libkgrpc.grpc_slice_from_copied_buffer
+import libkgrpc.grpc_slice_from_copied_string
+import libkgrpc.grpc_slice_ref
+import libkgrpc.grpc_slice_unref
+import libkgrpc.kgrpc_metadata_array_append
+import kotlin.experimental.ExperimentalNativeApi
+
+private value class GrpcKey private constructor(val name: String) {
```
Mr3zee: Probably better to call it `GrpcMetadataKey` even if it is private

grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/GrpcMetadata.jvm.kt
```diff
@@ -4,7 +4,63 @@
 
 package kotlinx.rpc.grpc
 
+import io.grpc.Metadata
 import kotlinx.rpc.internal.utils.InternalRpcApi
 
 @InternalRpcApi
 public actual typealias GrpcMetadata = io.grpc.Metadata
+
+public actual operator fun GrpcMetadata.get(key: String): String? {
```
Mr3zee, line 53: Does java lib perform ascii validation?

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcMetadata.kt
Jozott00: Currently there is no API that takes a `GrpcMetadata` object. Users can only manipulate the provided object in the interceptors. So currently, there is no use case of constructing a `GrpcMetadata` object. Should I still add it?

grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/GrpcMetadata.jvm.kt
Jozott00: Yes, they have the same checks as we do.

Jozott00:
@Mr3zee I think I misunderstood the outcome of our last discussion about the serialization API for metadata. My understanding was, that we drop the API, as a full implementation would require differentiation between binary serialization and ascii serialization. So we leave it up to the user who can just call the upcoming `codec()` function to encode/decode the object. 

But I am also fine with adding API for binary serialization only. In that case, should we call them `get(key: Key)` or `getBinary(key: Key)` to reflect that the serialization is binary?

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/GrpcMetadata.kt
Mr3zee: Oh, in that case no, it is fine

Issue #523, opened 2025-10-27
Method descriptor lies about return type if return type is Flow


rnett:
**Describe the bug**

**To Reproduce**

Kotlin 2.0.21, kotlin-rpc 0.10.1

Create a descriptor for a method that returns `Flow<...>`

**Expected behavior**

The descriptor's `returnType.kType` is `Flow<...>`

**Actual behavior**

`returnType.kType` is just the flow's type argument.

**Additional context**

If this needs to be kept for internal reasons, a `isFlow` property on the `RpcType` or method descriptor would be helpful.  `isNonSuspendFunction` seems to provide the same value, but doesn't necessarily have the same contract.

PR #526, opened 2025-11-04
Preserve Flow<...> in returnType and add returnsFlow flag on RpcCallable

MiletaA:
**Subsystem**
Compiler Plugin, Descriptor Core

**Problem Description**
Descriptors for RPC methods that return `Flow<T>` were exposing `returnType.kType == T`  instead of the declared `Flow<T>`.
It is issue #523 

**Solution**
Stop unwrapping `Flow<T>` to `T` in `RpcStubGenerator`, so the generated descriptor now keeps `returnType.kType == Flow<T>`. 
Add `val returnsFlow: Boolean` to `RpcCallable` and implement it in `RpcCallableDefault`; the compiler plugin sets this flag whenever the callable’s return type is a `Flow<*>`.

### DIFF
--- a/compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcStubGenerator.kt
+++ b/compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcStubGenerator.kt
@@ -797,7 +797,7 @@ internal class RpcStubGenerator(
                 type = ctx.rpcCallable.typeWith(declaration.serviceType),
                 symbol = ctx.rpcCallableDefault.constructors.single(),
                 typeArgumentsCount = 1,
-                valueArgumentsCount = 5,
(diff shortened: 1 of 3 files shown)


Mr3zee:
@MiletaA thank you for the contribution! 
This is good contribution in terms how you handled compiler code and tests. However I want to point out some problems with it:
- It is a breaking behaviour for kRPC and tests will fail
- Flow as a type doesn't have a lot of sense from the API perspective, it's better to use typed information. You can check out how we did it in a branch already: https://github.com/Kotlin/kotlinx-rpc/blob/cda97dc911441a9451a110df86d7159b7092b96a/core/src/commonMain/kotlin/kotlinx/rpc/descriptor/RpcServiceDescriptor.kt#L72

Mr3zee:
@MiletaA no, this is unrelated part of another feature. I was pointing out to the particular code part, not a commit. To backport it you'll need to extract the change in the descriptor from others in the branch.

PR #527, opened 2025-11-04
grpc: Add gzip Compression Support

Jozott00:
**Subsystem**
gRPC

**Problem Description**
The current API doesn't provide a way to set a compression algorithm for message encoding.

**Solution**
This PR adds support for gzip compression of messages for the client by adding a respective call option.


grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/internal/GrpcCallOptions.kt
Jozott00, line 1: Actually moved to the parent package (not deleted)

Mr3zee (approved):
Amazing work!

PR #528, opened 2025-11-05
grpc: Add keepalive Support

Jozott00:
**Subsystem**
gRPC

**Problem Description**
To reduce latency because of required reconnections to the server, the client sometimes wants to keep a connection open.

**Solution**
This can be done by setting the keepalive configuration option on the gRPC ClientConfiguration.

```kt
GrpcClient("localhost", 50051) {
    keepAlive {
         time = 10.seconds
         timeout = 20.seconds
         withoutCalls = false
    }
}
```

It also removes some tests that are either duplicates or testing undefined behavior (throwing in onClose of serverInterceptor).

### DIFF
--- a/grpc/grpc-client/src/nativeMain/kotlin/kotlinx/rpc/grpc/client/internal/ManagedChannel.native.kt
+++ b/grpc/grpc-client/src/nativeMain/kotlin/kotlinx/rpc/grpc/client/internal/ManagedChannel.native.kt
@@ -68,7 +68,6 @@ public actual fun ManagedChannelBuilder(target: String, credentials: ClientCrede
     return NativeManagedChannelBuilder(target, credentials)
 }
 
-
 internal actual fun ManagedChannelBuilder<*>.applyConfig(config: GrpcClientConfiguration): ManagedChannelBuilder<*> {
     this.config = config
     return this


grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/GrpcClient.kt
```diff
@@ -294,4 +294,65 @@ public class GrpcClientConfiguration internal constructor() {
...
+     * Client authors must coordinate with service owners for whether a particular client-side
+     * setting is acceptable.
+     *
+     * @property time Specifies the maximum amount of time the channel can remain idle before a
+     * keep-alive ping is sent to the server to check the connection state.
+     * The default value is `Duration.INFINITE`, which disables keep-alive pings when idle.
+     *
+     * @property timeout Sets the amount of time to wait for a keep-alive ping response.
+     * If the server does not respond within this timeout, the connection will be considered broken.
+     * The default value is 20 seconds.
```
Mr3zee, line 345: Any particular reason for 20 seconds?

Mr3zee (approved):
Great job!

grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/GrpcClient.kt
Jozott00: 20 seconds is defined by the gRPC specification https://grpc.io/docs/guides/keepalive/#keepalive-configuration-specification

PR #530, opened 2025-11-11
grpc: Call credentials

Jozott00:
**Subsystem**
gRPC

**Problem Description**
Until now, call credentials were missing in the gRPC client implementation.

**Solution**
This PR adds support for call credentials to the client, which can be used to provide application layer authentication and authorization.

### DIFF
--- a/grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/credentials.kt
+++ b/grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/credentials.kt
@@ -107,12 +107,12 @@ public class GrpcInsecureClientCredentials : GrpcClientCredentials
  * Verify the server's identity using a trust manager with root certificates:
  *
  * ```kotlin
- * val credentials = GrpcTlsClientCredentials {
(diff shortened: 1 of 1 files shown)


grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/credentials.kt
```diff
@@ -4,30 +4,244 @@
...
+ *
+ * Verify the server's identity using a trust manager with root certificates:
+ *
+ * ```kotlin
+ * val credentials = GrpcTlsClientCredentials {
+ *     trustManager(serverCertPem)
+ * }
+ *
+ * val client = GrpcClient("example.com", 443) {
+ *     credentials = credentials
```
Mr3zee: should be `this.credentials = credentials`

Mr3zee (approved):
Splendid!

grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/credentials.kt
Jozott00: good catch!

PR #534, opened 2025-11-19
Added full KMP support for the Gradle Plugin

Mr3zee:
**Subsystem**
Gradle Plugin

**Problem Description**
KMP support was working only for commonMain/Test sets

**Solution**
Add full KMP sources hierarchy support

### DIFF
--- a/gradle-plugin/src/main/kotlin/kotlinx/rpc/buf/tasks/BufExecTask.kt
+++ b/gradle-plugin/src/main/kotlin/kotlinx/rpc/buf/tasks/BufExecTask.kt
@@ -53,16 +53,41 @@ public abstract class BufExecTask @Inject constructor(
     @get:Input
     internal abstract val debug: Property<Boolean>
 
+    /**
+     * Properties of the buf task.
+     *
+     * Can be used with [BufTasks] to filter tasks.
+     */
     public open class Properties internal constructor(
+        /**
+         * Whether the task is for a test source set.
+         */
         public val isTest: Boolean,
+        /**
+         * Name of the [kotlinx.rpc.protoc.ProtoSourceSet] this task is associated with.
+         */
         public val sourceSetName: String,
     )
 
+    /**
+     * Properties of the buf task for android source sets.
+     *
+     * Can be used with [BufTasks] to filter tasks.
+     */
     public class AndroidProperties internal constructor(
         isTest: Boolean,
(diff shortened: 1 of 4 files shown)


gradle-plugin/src/main/kotlin/kotlinx/rpc/buf/tasks/BufExecTask.kt
```diff
@@ -27,22 +27,70 @@ import kotlinx.rpc.buf.BUF_GEN_YAML
...
+    @Internal
+    public val properties: Properties
+) : DefaultTask() {
     init {
         group = PROTO_GROUP
     }
 
+    // list of buf task dependencies of the same type
+    @get:Internal
+    internal abstract val bufTaskDependencies: SetProperty<String>
```
Jozott00, line 48: Is this used?

gradle-plugin/src/main/kotlin/kotlinx/rpc/Extensions.kt
```diff
@@ -48,29 +48,64 @@ public open class RpcExtension @Inject constructor(objects: ObjectFactory, priva
...
+    }
+
+    internal val protocApplied = AtomicBoolean(false)
+
+    internal val protocInternal by lazy {
+        if (protocApplied.get()) {
+            error("Illegal access to protoc extension during DefaultProtocExtension.init")
+        }
+
+        protocApplied.set(true)
```
Jozott00, line 90: I think this should be
```
if (protocApplied.getAndSet(true) { ... }
```
to make it actually thread safe

gradle-plugin/src/main/kotlin/kotlinx/rpc/buf/tasks/BufExecTask.kt
Mr3zee: Yes, in `bufDependsOn` implementation

gradle-plugin/src/main/kotlin/kotlinx/rpc/Extensions.kt
Mr3zee: yeah, you're right

Mr3zee: Fixed in another branch

PR #536, opened 2025-11-24
grpc: Status exception API refactoring

Jozott00:
**Subsystem**
gRPC

**Problem Description**
The current `StatusException` API is far from Kotlin idiomatic.

**Solution**
Because `StatusException` is tightly coupled with the Java `StatusException` as it is just a typealias, we have some constraints in regard to changes of the implementation.

This PR tries to address the most inconvenient parts of the current API by using extension functions:
- `StatusException.status`
- `StatusException.trailers`
- `Status.description`
- `Status.statusCode`
- `StatusCode.asException`
- `StatusCode.asStatus`
- `Status.asException`

### DIFF
--- a/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/Status.kt
+++ b/grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/Status.kt
@@ -6,6 +6,8 @@
 
 package kotlinx.rpc.grpc
 
+import kotlinx.rpc.internal.utils.InternalRpcApi
+
 /**
  * Defines the status of an operation by providing a standard [StatusCode] in conjunction with an
  * optional descriptive message.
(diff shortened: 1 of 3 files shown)


grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/Status.kt
```diff
@@ -26,14 +28,39 @@ package kotlinx.rpc.grpc
...
+ * The status code of this status.
+ */
 public expect val Status.statusCode: StatusCode
 
+/**
+ * The description of this status, or null if not present.
+ */
+public val Status.description: String? get() = getDescription()
+
+// this is currently @InternalRpcApi as it's behavior would be inconsistent between JVM and Native.
```
Mr3zee, line 50: what's the inconsistency?

Mr3zee (approved):
lgtm

grpc/grpc-core/src/commonMain/kotlin/kotlinx/rpc/grpc/Status.kt
Jozott00: Some APIs in Java attach the causing exception. However, on native, we would lose the causing exception information as we can't pass it through the C layer. So I think it might be better to just omit it, so users don't rely on the cause value of the Status exception.

PR #539, opened 2025-11-27
pb: Add `copy` method to generated messages

Jozott00:
**Subsystem**
Protobuf

**Problem Description**
Messages had no `copy` method to easily creating an altered message instance.

**Solution**
This PR adds the `copy {}` extension function, which will produce a deep copy of the receiver message.
Users can override copied fields as with the `invoke`/constructor function.

**Limitation**
The copy function should also preserve all unknown message fields.
However, as unknown fields are not yet stored (just discarded), this isn't implemented yet.

### DIFF
--- a/protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
+++ b/protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
@@ -387,20 +387,23 @@ class ModelToProtobufKotlinCommonGenerator(
             // e.g., internal map entries don't need a copy() method
             return
         }
-        val demoField = declaration.actualFields.firstOrNull()?.name ?: "someField"
+        val demoField = declaration.actualFields.firstOrNull()?.name
+        val invocation = if (demoField == null) "()" else """
+            | {
+            |    $demoField = ...    
+            |}
+        """.trimMargin()
         function(
             name = "copy",
             contextReceiver = declaration.name.safeFullName(),
             args = "body: ${declaration.internalClassFullName()}.() -> Unit = {}",
             returnType = declaration.name.safeFullName(),
             comment = Comment.leading("""
-                Copies the original message, including unknown fields.
-                ```   
-                val copy = original.copy {
-                    $demoField = ...    
-                }
-                ```
-            """.trimIndent())
+                |Copies the original message, including unknown fields.
+                |```
+                |val copy = original.copy$invocation
+                |```
+            """.trimMargin())
         ) {
             code("return this.asInternal().copyInternal(body)")
         }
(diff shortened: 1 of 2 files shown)


protobuf/protobuf-core/src/commonMain/generated-code/kotlin-multiplatform/com/google/protobuf/kotlin/_rpc_internal/Api.kt
```diff
@@ -76,6 +76,22 @@ public class ApiInternal: com.google.protobuf.kotlin.Api, kotlinx.rpc.protobuf.i
         }
     }
 
+    public override fun copy(body: ApiInternal.() -> Unit): ApiInternal { 
+        val copy = ApiInternal()
+        copy.name = name
+        copy.methods = methods.map { it.copy() }
+        copy.options = options.map { it.copy() }
+        copy.version = version
+        if (presenceMask[0]) {
```
Mr3zee, line 86: Should we also copy presence masks?

protobuf/protobuf-core/src/commonMain/generated-code/kotlin-multiplatform/com/google/protobuf/kotlin/Any.kt
```diff
@@ -128,5 +128,10 @@ public interface Any {
     */
     public val value: ByteArray
 
+    /**
+    * Copies the original message, including unknown fields.
+    */
+    public fun copy(body: com.google.protobuf.kotlin.AnyInternal.() -> Unit = {}): com.google.protobuf.kotlin.Any
```
Mr3zee: Let's make it an extension function and temporary put it in __rpc_internal directory.

I want to keep the original messages as small and as similar to proto as possible. Later we may create a distinction: public, public extensions, internal  - generated code types

protobuf/protobuf-core/src/commonMain/generated-code/kotlin-multiplatform/com/google/protobuf/kotlin/_rpc_internal/Api.kt
Jozott00: The mask is implicitly constructed by setting the copied values.

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -474,12 +589,21 @@ class ModelToProtobufKotlinCommonGenerator(
     private fun CodeGenerator.generateMessageConstructor(declaration: MessageDeclaration) {
         if (!declaration.isUserFacing) return
 
+        val demoField = declaration.actualFields.firstOrNull()?.name ?: "someField"
```
Mr3zee: someField may be confusing for messages with no fields, let's just omit the fields entirely in this case

Jozott00: I tried that, but for some reason indentation was not working properly xD I will try again.

PR #540, opened 2025-12-01
pb: Add presence check support

Jozott00:
**Subsystem**
Protobuf

**Problem Description**
Even if a field in a message is optional, we can't make it nullable. This is because fields may either have a default value, or we want to be able to access a default value in a nested field such that the parent object must not be null.

However, users must be able to distinguish between a field that was manually set with the default value and one that was not actively set.

**Solution**
This PR solves this problem by adding a `presence` API to each message that contains fields with presence tracking.
```kt
val myMessage = Message { myField = "value" }
myMessage.presence.hasField
```
`presence` is an extension getter on the generated message interface that provides a `hasXYZ` getter for all fields with tracked/explicit presence. 
This intermediate `presence` object is employed to ensure concise code completion and minimize the number of suggestions provided by the IDE.

The only field type that lacks a `hasXYZ` getter while its presence is tracked is the `oneof` field. Since `oneof` fields are nullable, they are null if they are not set.

### DIFF
--- a/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf/conformance/Conformance.ext.kt
+++ b/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf/conformance/Conformance.ext.kt
@@ -1,0 +1,148 @@
+@file:OptIn(ExperimentalRpcApi::class, InternalRpcApi::class)
+package com.google.protobuf.conformance
+
+import kotlin.jvm.JvmInline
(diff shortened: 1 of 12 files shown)


Jozott00:
@Mr3zee 
1. Yes, that is probably better. Any suggestions on how to name the interface? Currently it is an inner class which makes it easy. When moving it to the `.ext.kt` should I name the interface `<OuterClass><InnerClass>Presence`? (assuming it's the interface for a nested message)
2. Good point, I will add documentation

PR #545, opened 2025-12-15
grpc: Fix `retry/retryWhen` logic on rpc flow

Jozott00:
**Subsystem**
gRPC

**Problem Description**
The current implementation results in unexpected behavior when users call the `retry/retryWhen` API on the flow returned by a server streaming RPC service. 
```kt
myService.receiveMessages(user).retryWhen { cause, _ ->
   // retry after 2000 if server was not available
   val shouldRetry = (cause as? StatusException)?.status?.statusCode != StatusCode.UNAVAILABLE
   if (shouldRetry) delay(2000)
   shouldRetry
}
```
While one would expect that the RPC call is retried and might eventually succeed,
it will always cause an exception (even if the retried call succeeded)

**Solution**
The cause for the problem lies in the flow creation. The flow creation happens after interceptors were applied to the request and after certain states in the call scope were created.

To fix this, we wrap the call scope creation in a `flow{}`, so if the user retries this flow, it will also create a new call scope object.


Mr3zee (approved):
lgtm

PR #555, opened 2026-01-14
Hide compiler generated code

Mr3zee:
**Subsystem**
Compiler Plugin

**Problem Description**
Generated `$rpcServiceStub` is shown in the UI

**Solution**
Hide it


compiler-plugin/compiler-plugin-k2/src/main/kotlin/kotlinx/rpc/codegen/FirRpcUtils.kt
```diff
@@ -51,3 +66,30 @@ fun List<FirAnnotation>.rpcAnnotation(session: FirSession, predicate: Declaratio
...
+
+// stolen from kx.serialization
+fun createDeprecatedHiddenAnnotation(session: FirSession): FirAnnotation = buildAnnotation {
+    val deprecatedAnno = session.symbolProvider
+        .getClassLikeSymbolByClassId(StandardClassIds.Annotations.Deprecated) as FirRegularClassSymbol
+
+    annotationTypeRef = vsApi { deprecatedAnno.defaultType().toFirResolvedTypeRefVS() }
+
+    argumentMapping = buildAnnotationArgumentMapping {
+        mapping[Name.identifier("message")] = buildLiteralExpression(
```
e5l: could be constant `Name.identifier("message")`

PR #556, opened 2026-01-14
Release 0.10.2

Mr3zee:
**Release**

* Kotlin 2.3.0
* Kotlin 2.3.20-Beta1
* Removed internal generated declarations from UI

### DIFF
--- a/versions-root/libs.versions.toml
+++ b/versions-root/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 # core library version
-kotlinx-rpc = "0.11.0-SNAPSHOT"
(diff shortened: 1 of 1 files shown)


README.md
lukewarlow, line 106: Presumably this should be updated too?

Mr3zee: It is fine, isn't it?

Mr3zee: If you mean Kotlin 2.3.20-Beta1, then we don't write unstable versions because their support is not guaranteed

lukewarlow: 🤦🏼‍♂️ I misread the two versions, sorry for the noise.

PR #557, opened 2026-01-15
pb: Support unknown field propagation

Jozott00:
**Subsystem**
Protobuf

**Problem Description**
Currently, all fields that are not defined in the proto file will be ignored.
However, protobuf libraries must be able to preserve them within the message and encode them if the message is serialized again.

**Solution**
This PR implements unknown field preservation and propagation by storing unknown fields read during decoding into a buffer. During encoding this buffer is emitted together with the rest of the message. If a message is copied, the unknown field buffer is also copied.

This change does not affect the current message API, so there is no way to access a message's unknown fields programmatically. This is similar to some other implementations, such as Go's one. 

A future PR will also provide an API to configure if unknown fields should be discarded.

### DIFF
--- a/protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/codeRequestToModel.kt
+++ b/protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/codeRequestToModel.kt
@@ -92,7 +92,7 @@ fun Descriptors.GenericDescriptor.fqName(): FqName {
         is Descriptors.EnumDescriptor -> FqName.Declaration(nameCapital, containingType?.fqName() ?: file.fqName())
         is Descriptors.EnumValueDescriptor -> FqName.Declaration(name, type.fqName())
         is Descriptors.ServiceDescriptor -> FqName.Declaration(nameCapital, file.fqName())
-        is Descriptors.MethodDescriptor -> FqName.Declaration(nameLower, service.fqName())
+        is Descriptors.MethodDescriptor -> FqName.Declaration(nameLower, service?.fqName() ?: file.fqName())
         else -> error("Unknown generic descriptor: $this")
     }
     nameCache[this] = fqName
--- a/protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/model/FqName.kt
+++ b/protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/model/FqName.kt
@@ -76,8 +76,8 @@ fun FqName.packageName(): FqName.Package {
     }
 }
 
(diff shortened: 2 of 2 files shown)


protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -745,8 +802,12 @@ class ModelToProtobufKotlinCommonGenerator(
                                 code("throw $PB_PKG.ProtobufDecodingException(\"Unexpected END_GROUP tag.\")")
                             })
                         }
-                        code("// we are currently just skipping unknown fields (KRPC-191)")
-                        code("decoder.skipValue(tag)")
+
+                        // TODO: Use configuration to disable it
```
Mr3zee, line 806: Please make a YT ticket (or link existing) for the TODO

tests/protobuf-conformance/src/test/kotlin/kotlinx/rpc/protoc/gen/test/ConformanceTest.kt
```diff
@@ -21,19 +21,9 @@ import kotlin.io.path.readLines
...
-    "Required.Editions_Proto3.ProtobufInput.UnknownOrdering.ProtobufOutput",
-    "Required.Editions_Proto3.ProtobufInput.UnknownVarint.ProtobufOutput",
     "Required.Proto2.ProtobufInput.MessageSetEncoding.UnknownExtension.ProtobufOutput",
-    "Required.Proto2.ProtobufInput.UnknownOrdering.ProtobufOutput",
-    "Required.Proto2.ProtobufInput.UnknownVarint.ProtobufOutput",
     "Required.Proto2.ProtobufInput.ValidMessageSetEncoding.OutOfOrderGroupsEntries.ProtobufOutput",
     "Required.Proto2.ProtobufInput.ValidMessageSetEncoding.ProtobufOutput",
-    "Required.Proto3.ProtobufInput.UnknownOrdering.ProtobufOutput",
-    "Required.Proto3.ProtobufInput.UnknownVarint.ProtobufOutput",
 )
```
Mr3zee, line 27: Nice!!!

Mr3zee (approved):
Great work, thanks!

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: This is already implemented in #560

PR #560, opened 2026-01-27
grpc+pb: Add codec configuration

Jozott00:
**Subsystem**
gRPC + Protobuf

**Problem Description**
Currently, it is not possible to set a `MessageCodec` configuration to dynamically influence the codec behavior during runtime. However, some codecs would benefit from such a feature, e.g. for the generated protobuf codecs it should be possible to turn of unknown field preservation.

Additionally, we need a way to easily receive the `MessageCodec` from a class annotated with `@WithCodec`. 

**Solution**
This PR adds support for both. It is now possible to retrieve the codec with
```kt
@WithCodec(MyMessageCodec::class)
class MyMessage
object MyMessageCodec : MessageCodec<MyMessage> { ... }

val codec = codec<MyMessage>()
```
The implementation does not require additional IR compiler plugin code, as we are using the `@AssociatedObjectKey` feature to attach the codec to the message type on native targets.

The `codec` function also takes a `CodecConfig` (which is just a marker interface), and returns a `MessageCodec` that wraps the actual one, to pass the given default config for every encode/decode call.

```kt
```

The `CodecConfig` can also be set in the `GrpcClientConfiguration` and `GrpcServerConfiguration`. This will ensure that the configuration is passed to the `MessageCodec` for each message encoding or decoding. 
To implement this, I had to adjust compiler plugin IR generation a bit to check if the set `CodecConfig` is null or not.

---

**ServiceName overwrite**: Another mini feature that I required for testing was to be able to set the `protoServiceName` in the `@Grpc` annotation. As the `protoPackage`, it allows the user to overwrite the service name used by the gRPC protocol.
This allowed me to write a test service like the following
```kt
@Grpc(protoServiceName = "Service")
private interface ClientService {
    suspend fun serverTest(message: UnknownFieldsAll): UnknownFieldsAll
    suspend fun clientTest(message: UnknownFieldsAll): UnknownFieldsSubset
}

@Grpc(protoServiceName = "Service")
private interface ServerService {
    suspend fun serverTest(message: UnknownFieldsSubset): UnknownFieldsSubset
    suspend fun clientTest(message: UnknownFieldsAll): UnknownFieldsAll
}
```
It might also be necessary if, in edge cases, the generated service interfaces should have a different name than the name specified in the proto file.

---

**CheckedTypeAnnotation customization**: Another feature required for implementing the `codec()` function, was to add a new optional parameter to the `@CheckedTypeAnnotation` annotation, which now also has a 
```kt
val checkFor: KClass<*> = CheckedTypeAnnotation::class
```
parameter. This is the default value, leading to the same behavior as before.
When this parameter is set, the FIR compiler check no longer verifies that the type
is annotated with `@CheckedTypeAnnotation` annotated annotation itself (as is the case for `@Rpc` and `@Grpc`).
Instead, it validates that the type is annotated with the annotation specified
by `checkFor`.

This is required because the `@WithCodec` annotation takes an argument, so we can't use it like `fun codec<reified @WithCodec T : Any>()`. Instead, I defined a new annotation
```kt
@InternalRpcApi
@CheckedTypeAnnotation(WithCodec::class)
@Target(AnnotationTarget.TYPE_PARAMETER)
public annotation class HasWithCodec
```
which allows us to define the `codec` function as
```kt
fun codec<reified @HasWithCodec T : Any>()
```

---

I know that this is a huge PR with many changes and features. If necessary, I can split it up into smaller ones. 🤓

Are you ok with the proposed solutions or would you prefer something else?

### DIFF
--- a/tests/compiler-plugin-tests/src/testData/diagnostics/withCodec.fir.txt
+++ b/tests/compiler-plugin-tests/src/testData/diagnostics/withCodec.fir.txt
@@ -160,3 +160,13 @@ FILE: module_main_withCodec.kt
         }
 
     }
+    public final class Test10 : R|kotlin/Any| {
+        public constructor(): R|Test10| {
+            super<R|kotlin/Any|>()
+        }
+
+    }
+    public final val testCodec10: R|kotlinx/rpc/grpc/codec/MessageCodec<Test10>| = R|kotlinx/rpc/grpc/codec/codec|<R|Test10|>()
+        public get(): R|kotlinx/rpc/grpc/codec/MessageCodec<Test10>|
+    public final val testCodec: R|kotlinx/rpc/grpc/codec/MessageCodec<Test>| = R|kotlinx/rpc/grpc/codec/codec|<R|Test|>()
+        public get(): R|kotlinx/rpc/grpc/codec/MessageCodec<Test>|
--- a/tests/compiler-plugin-tests/src/testData/diagnostics/withCodec.kt
+++ b/tests/compiler-plugin-tests/src/testData/diagnostics/withCodec.kt
@@ -11,6 +11,7 @@
 import kotlinx.rpc.protobuf.input.stream.InputStream
 import kotlinx.rpc.grpc.codec.CodecConfig
 import kotlinx.rpc.grpc.codec.WithCodec
+import kotlinx.rpc.grpc.codec.codec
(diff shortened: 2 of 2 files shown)


Jozott00:
Close in favor of #563, #564, #565 and #569 .

PR #565, opened 2026-02-04
grpc: Add `CodecConfig` and `codec` function

Jozott00:
**Subsystem**
gRPC

**Problem Description**
Currently, it is not possible to set a MessageCodec configuration to dynamically influence the codec behavior during runtime. However, some codecs would benefit from such a feature, e.g. for the generated protobuf codecs it should be possible to turn of unknown field preservation.

Additionally, we need a way to easily receive the MessageCodec from a class annotated with @WithCodec.

**Solution**
This PR adds support for both. It is now possible to retrieve the codec with
```kt
@WithCodec(MyMessageCodec::class)
class MyMessage
object MyMessageCodec : MessageCodec<MyMessage> { ... }

val codec = codec<MyMessage>()
```
The implementation does not require additional IR compiler plugin code, as we are using the `@AssociatedObjectKey` feature to attach the codec to the message type on native targets.

The `codec` function also takes a `CodecConfig` (which is just a marker interface), and returns a `MessageCodec` that wraps the actual one, to pass the given default config for every encode/decode call.

```kt
```

The `CodecConfig` can also be set in the `GrpcClientConfiguration` and `GrpcServerConfiguration`. This will ensure that the configuration is passed to the `MessageCodec` for each message encoding or decoding. 
To implement this, I had to adjust compiler plugin IR generation a bit to check if the set `CodecConfig` is null or not.


grpc/grpc-codec/src/commonMain/kotlin/kotlinx/rpc/grpc/codec/WithCodec.kt
```diff
@@ -4,9 +4,134 @@
...
+ * Specifies a custom [MessageCodec] for encoding and decoding a message type in gRPC communication.
+ *
+ * If a gRPC service uses a message type annotated with this annotation,
+ * the generated gRPC stubs will use the specified codec to encode and decode instances of that type.
+ *
+ * The [MessageCodec] for a type annotated with [WithCodec] can be retrieved using the [codec] function.
+ *
+ * Example:
+ * ```
+ * @WithCodec(ChatEntry::class)
```
Mr3zee: `@WithCodec(ChatEntryCodec::class)`

grpc/grpc-codec/src/commonMain/kotlin/kotlinx/rpc/grpc/codec/WithCodec.kt
```diff
@@ -4,9 +4,134 @@
...
+ * @param T The message type for which to retrieve the codec. Must be annotated with [WithCodec].
+ * @param kClass The [KClass] of the message type ([T]) for which to retrieve the codec.
+ * @param config Optional default [CodecConfig] to use for encoding and decoding operations.
+ * @return A [MessageCodec] instance for the specified class.
+ *
+ * @see WithCodec
+ * @see MessageCodec
+ * @see CodecConfig
+ */
+public fun <@HasWithCodec T: Any> codec(kClass: KClass<T>, config: CodecConfig? = null): MessageCodec<T> {
```
Mr3zee: Please, make two `codec` functions have the same docs (except for parameters), and please add a function that accepts `KType` instead of `KClass`

grpc/grpc-codec/src/commonMain/kotlin/kotlinx/rpc/grpc/codec/WithCodec.kt
```diff
@@ -4,9 +4,134 @@
...
+    val codecObj = resolveCodec(kClass) ?: error("No codec object found for ${kClass.qualifiedName}. " +
+            "Make sure that the kotlinx.rpc compiler plugin is applied.")
+    @Suppress("UNCHECKED_CAST")
+    val codec = codecObj as? MessageCodec<T> ?: error("Internal kotlinx.rpc error: " +
+            "Codec for ${kClass.qualifiedName} is not a MessageCodec but ${codecObj::class.qualifiedName}")
```
Mr3zee: Please, use `MessageCodec::class.simpleName` for future refectorings

grpc/grpc-codec/src/commonTest/kotlin/kotlinx/rpc/grpc/codec/test/CodecTest.kt
```diff
@@ -0,0 +1,80 @@
...
+
+
+@WithCodec(MyMessageCodec::class)
+data class MyMessage(
+    val value: String
+)
+
+class MyCodecConfig(
+    val appendHello: Boolean
+): CodecConfig
```
Mr3zee, line 27: Add a nl, please

grpc/grpc-codec/src/commonMain/kotlin/kotlinx/rpc/grpc/codec/internal/ConfiguredMessageCodecDelegate.kt
```diff
@@ -0,0 +1,33 @@
...
+import kotlinx.rpc.protobuf.input.stream.InputStream
+
+/**
+ * A [MessageCodec] that delegates encoding and decoding to another codec,
+ * but with a configured default [CodecConfig].
+ */
+// Used in the compiler plugin when creating/resolving the message codec in the methodDescriptor, to
+// return construct a message codec that wraps the actual message codec but applies the default config
+// given during client configuration.
+@InternalRpcApi
```
Mr3zee, line 20: Should probably be as one comment

grpc/grpc-codec/src/commonMain/kotlin/kotlinx/rpc/grpc/codec/WithCodec.kt
```diff
@@ -4,9 +4,134 @@
...
+
+    if (config == null) {
+        // if no default config is specified, we just return the plain codec
+        return codec
+    }
+
+    // otherwise, we wrap the codec to pass the default config on every encode/decode
+    val defaultConfig = config
+    return object : MessageCodec<T> {
```
Mr3zee: Can you use `ConfiguredMessageCodecDelegate` here?

Mr3zee (approved):
lgtm!
please run compiler tests, and see comments below

grpc/grpc-codec/src/commonMain/kotlin/kotlinx/rpc/grpc/codec/WithCodec.kt
Jozott00: Good catch! I now moved the `ConfiguredMessageCodecDelegate` into the `grpc-codec` module and now also use it here.

PR #566, opened 2026-02-07
pb: Fix size discrepancy

### LINKED ISSUE #561: Illegal Wire Type 4 (EndGroup) appended to generated gRPC response messages
I am encountering a protocol violation where the server appends an unexpected tag to the end of a gRPC response. Specifically, after the expected fields of the response message are serialized, the stream contains an additional Field Number 12 with Wire Type 4 (EndGroup).

This causes standard Protobuf parsers and network inspectors to fail with an "Unknown or unsupported wiretype" error. Wire Type 4 is deprecated and must be preceded by a matching StartGroup (Wire Type 3); its presence here as a standalone "footer" violates the Proto3 wire specification and breaks compatibility with non-Kotlin gRPC clients.

## Reproduction Structure
The issue occurs with service interfaces and messages generated automatically from .proto files.

### Sample .proto Definition

```proto
syntax = "proto3";

package com.example.generic;

message ChildItem {
    int64 timestamp = 1;
    double value = 2;
}

message RootResponse {
    repeated ChildItem items = 1;
    string metadata = 2;
}

service ExampleService {
    rpc GetData (EmptyRequest) returns (RootResponse);
}
```

## Evidence (Network Analysis)
Below is a breakdown of the response stream. The parser correctly iterates through the generated message fields, but fails immediately after the last valid field of the root message.

**Binary Breakdown of the Malformed Footer:** `[Hex: 64]` -> `[Binary: 01100 100]` -> `[Field Number: 12, Wire Type: 4 (End Group)]`

### Inspector Log (Genericized)

```text
Protocol Buffers: /com.example.generic.ExampleService/GetData, response
  Message: com.example.generic.RootResponse
    Field(1): items (message)
      Value Length: 280
      items: (280 bytes)
        Message: com.example.generic.ChildItem
          Field(1): timestamp = 1738368000000 (int64)
          Field(2): value = 309883250.0 (double)
    Field(2): metadata = "sample_meta" (string)
    
    Field(12): <UNKNOWN>
      .110 0... = Field Number: 12
      .... .100 = Wire Type: End group (deprecated) (4)
      [Expert Info (Warning/Protocol): Unknown or unsupported wiretype]
```

## Debugging and Maintenance Barriers
Investigating this issue is currently difficult due to the project's infrastructure and release workflow:

*   **Lack of Snapshot/Nightly Releases:** There are no automated builds for grpc-common changes. I am forced to build the project locally and publish to `mavenLocal` to debug. This makes it nearly impossible to provide a shareable, isolated reproduction project, as the required dependencies are not available in any public repository.
*   **Build System Portability:** The build process for `kotlinx-rpc` appears highly optimized for internal JetBrains environments. Several release tools trigger warnings regarding missing internal developer tokens, making it cumbersome for community members to contribute or verify fixes at the protocol level.

## Proposal: Automated Releases
To improve the feedback loop for wire-protocol issues, I strongly advise implementing an automated Nightly or Snapshot release mechanism for gRPC-related modules.

Moving away from manual pushes to specific `grpc-release` branches and providing public snapshots would:

1.  **Enable users to verify protocol-level fixes** in real-world applications without complex local build setups.
2.  **Make the project more accessible** to the community by removing the need to manage internal-leaning build environments.
3. **Reduce maintainer effort** to release intermediate release candidate versions
3.  **Allow for shareable reproductions** that point to public experimental artifacts.

fabiocarneiro:
Fixes #561 

This PR fixes a bug in the protobuf code generator where the calculated `_size` property of messages containing repeated fields or maps was incorrect. This discrepancy occurred because the generator failed to account for tag sizes and length-delimited prefixes (varint length) for elements within these collections.

Before this fix:

* **Maps:** Only the size of the internal entry message was being added, missing the tag and the length prefix for the entry itself.
* **Repeated Fields:** For `LENGTH_DELIMITED` types (strings, nested messages, bytes), the calculation only included the data size and tag, omitting the varint length prefix.

## Key Changes

* **Code Generation:** Modified `ModelToProtobufKotlinCommonGenerator.kt` to:
    * Use `int32SizeCall` to include the size of the length prefix for `LENGTH_DELIMITED` fields in maps and repeated collections.
    * Correctly handle `START_GROUP` and `END_GROUP` wire types (supporting legacy proto2 groups).
    * Ensure tag sizes are correctly added for all elements in a repeated field.
* **Well-Known Types:** Regenerated the checked-in well-known types (`Api.kt`, `Struct.kt`, `Type.kt`, etc.) to include the corrected size calculation logic.
* **Regression Tests:** Introduced `ProtobufSizeCalculationTest.kt` which compares the generated `._size` against the actual `encoded.size` for:
    * Repeated nested messages.
    * Repeated strings.
    * Maps with primitive keys and values.

### DIFF
--- a/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf/conformance/_rpc_internal/Conformance.kt
+++ b/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf/conformance/_rpc_internal/Conformance.kt
@@ -665,7 +665,7 @@ fun com.google.protobuf.conformance.FailureSetInternal.Companion.decodeWith(msg:
 private fun com.google.protobuf.conformance.FailureSetInternal.computeSize(): Int { 
     var __result = 0
     if (test.isNotEmpty()) { 
-        __result += test.sumOf { it.asInternal()._size + kotlinx.rpc.protobuf.internal.WireSize.tag(2, kotlinx.rpc.protobuf.internal.WireType.LENGTH_DELIMITED) }
+        __result += test.sumOf { it.asInternal()._size.let { kotlinx.rpc.protobuf.internal.WireSize.tag(2, kotlinx.rpc.protobuf.internal.WireType.LENGTH_DELIMITED) + kotlinx.rpc.protobuf.internal.WireSize.int32(it) + it } }
     }
 
     return __result
--- a/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf_test_messages/edition2023/_rpc_internal/TestMessagesEdition2023.kt
+++ b/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf_test_messages/edition2023/_rpc_internal/TestMessagesEdition2023.kt
@@ -4068,19 +4068,19 @@ private fun com.google.protobuf_test_messages.edition2023.TestAllTypesEdition202
     }
 
     if (repeatedString.isNotEmpty()) { 
-        __result += repeatedString.sumOf { kotlinx.rpc.protobuf.internal.WireSize.string(it) + kotlinx.rpc.protobuf.internal.WireSize.tag(44, kotlinx.rpc.protobuf.internal.WireType.LENGTH_DELIMITED) }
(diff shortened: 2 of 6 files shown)


Jozott00 (changes requested):
Thanks a lot for tackling the issue!

Please also run the `tests:protobuf-conformance:bufGenerateMain` task to regenerate protobuf conformance tests.

Apart from that, the PR looks very good to me.

Jozott00:
From my side this looks good now. I just merged #565, so a rebase and regeneration of the conformance tests and well-known types will be necessary (again).

@Mr3zee could you check if all changes are fine and handle the merge?

PR #569, opened 2026-02-09
pb: Add `ProtobufConfig` with `discardUnknownFields` configuration

Jozott00:
**Subsystem**
Protobuf

**Problem Description**
Currently, it is not possible to tell the protobuf decoder to discard unknown fields during message parsing. 

**Solution**
This PR adds the `ProtobufConfig` class that holds the `discardUnknownFields` property.
If passed to the `MessageCodec`, it tells the generated protobuf decoder to discard unknown fields.

```
```

### DIFF
--- a/grpc/grpc-core/src/desktopTest/kotlin/kotlinx/rpc/grpc/test/integration/GrpcCodecConfigTest.kt
+++ b/grpc/grpc-core/src/desktopTest/kotlin/kotlinx/rpc/grpc/test/integration/GrpcCodecConfigTest.kt
@@ -115,5 +115,4 @@ class GrpcCodecConfigTest : GrpcTestBase() {
         assertEquals(message, decoded)
     }
 
-
 }
--- a/protobuf/protobuf-core/src/commonMain/kotlin/kotlinx/rpc/protobuf/ProtobufConfig.kt
+++ b/protobuf/protobuf-core/src/commonMain/kotlin/kotlinx/rpc/protobuf/ProtobufConfig.kt
@@ -20,7 +20,7 @@ import kotlinx.rpc.grpc.codec.CodecConfig
  * val myCodec = codec<MyMessage>(config)
  *
  * // Or pass config per-operation
- * val decoded = codec.decode(stream, ProtobufConfig(discardUnknownFields = true))
+ * val decoded = codec.decode(stream, config)
  * ```
  *
  * @property discardUnknownFields When `true`, unknown fields encountered during deserialization


grpc/grpc-core/src/desktopTest/kotlin/kotlinx/rpc/grpc/test/integration/GrpcCodecConfigTest.kt
```diff
@@ -0,0 +1,119 @@
...
+        assertTrue(response.asInternal()._unknownFields.size > 0)
+
+        // encode and decode as UnknownFieldsAll
+        val codec = codec<UnknownFieldsAll>()
+        val decoded = codec.decode(codec.encode(message))
+        // should have preserved all fields
+        assertEquals(message, decoded)
+    }
+
+
```
Mr3zee: Redundant two nls

protobuf/protobuf-core/src/commonMain/kotlin/kotlinx/rpc/protobuf/ProtobufConfig.kt
```diff
@@ -0,0 +1,34 @@
...
+ * `.proto` files using `kotlinx-rpc`.
+ *
+ * Example:
+ * ```kotlin
+ * // Create a codec with custom config
+ * val config = ProtobufConfig(discardUnknownFields = true)
+ * val myCodec = codec<MyMessage>(config)
+ *
+ * // Or pass config per-operation
+ * val decoded = codec.decode(stream, ProtobufConfig(discardUnknownFields = true))
```
Mr3zee: Can be shorter with `codec.decode(stream, config)` (`config` is from the line above)

protobuf/protobuf-core/src/commonTest/kotlin/kotlinx/rpc/protobuf/test/ProtosTest.kt
```diff
@@ -374,9 +374,9 @@ class ProtosTest {
             arg3 = "fourth"
         }
 
-        encoder.writeMessage(1, firstPart as OtherInternal) { encodeWith(encoder) }
+        encoder.writeMessage(1, firstPart as OtherInternal) { encodeWith(encoder, null) }
```
Mr3zee, line 377: Should the parameter be with a default `null` value?

Mr3zee (approved):
LGTM

protobuf/protobuf-core/src/commonTest/kotlin/kotlinx/rpc/protobuf/test/ProtosTest.kt
Jozott00: I think it might be safer to leave it explicit. The API is only used by generated code (except for this test case) that must always propagate the configuration from its own arguments.

PR #571, opened 2026-02-10
grpc+pb: Remove InputStream abstraction in common

Jozott00:
**Subsystem**
Grpc+Protobuf

**Problem Description**
Currently we have a very thin user exposed abstraction layer of `InputStream` over `Buffer/Source` from `koltlinx.io`. This makes it cumbersome to use, as we use `Buffer` for everything, but always require to call `asInputStream()` or `asSource()` at every common-jvm boundary on the *common* side. 

Especially, users implementing their own `MessageCodec` must deal with it or use the `SourcedMessageCodec` which makes it confusing.

**Solution**
Previously the `Source` to `InputStream` conversion happened directly in the `MessageCodec` implementation:
```kotlin
public override fun encode(value: com.google.protobuf.kotlin.Any, config: CodecConfig?): InputStream { 
	val buffer = Buffer()
	val encoder = WireEncoder(buffer)
	// encoding ...
	return buffer.asInputStream()
}
```
On JVM the `expect class InputStream` was a typealias for `java.io.InputStream` and `asInputStream()` was defined as
```kotlin
public actual inline fun Source.asInputStream(): InputStream = asInputStream()
```
which is a `kotlin.io` extension function that returns a Java `InputStream` wrapping the `Source`. So there is no copying just delegation.

This conversion of `Source` to `InputStream` is now moved to the Marshaller of the gRPC JVM.
To make this possible, we had to remove the `Marshaller<T>` interface from the common expect class `MethodDescriptor`
```diff
public expect class MethodDescriptor<Request, Response> {
    public fun getFullMethodName(): String
    public fun getServiceName(): String?
-   public fun getRequestMarshaller(): Marshaller<Request>
-   public fun getResponseMarshaller(): Marshaller<Response>
    public fun getSchemaDescriptor(): Any?
    public fun isIdempotent(): Boolean
    public fun isSafe(): Boolean
    public fun isSampledToLocalTracing(): Boolean
-
-   public interface Marshaller<T> {
-       public fun stream(value: T): InputStream
-       public fun parse(stream: InputStream): T
-   }
}
```
The `private fun <T> MessageCodec<T>.toJvm()` now takes care of turning the `Source` into a input stream and vice-versa 
```kotlin
private fun <T> MessageCodec<T>.toJvm(): io.grpc.MethodDescriptor.Marshaller<T> {
    return object : io.grpc.MethodDescriptor.Marshaller<T> {
        override fun stream(value: T): InputStream {
            // wraps the source in a stream
            return encode(value).asInputStream()
        }

        override fun parse(stream: InputStream): T {
            // wraps the stream in a buffered source
            return decode(stream.asSource().buffered())
        }
    }
}
```

### DIFF
--- a/grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/descriptor/MethodDescriptor.jvm.kt
+++ b/grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/descriptor/MethodDescriptor.jvm.kt
@@ -31,7 +31,7 @@ internal val MethodType.asJvm: io.grpc.MethodDescriptor.MethodType
         MethodType.UNKNOWN -> io.grpc.MethodDescriptor.MethodType.UNKNOWN
     }
 
-private fun <T> MessageCodec<T>.toJvm(): io.grpc.MethodDescriptor.Marshaller<T> {
+private fun <T> MessageCodec<T>.toMarshaller(): io.grpc.MethodDescriptor.Marshaller<T> {
     return object : io.grpc.MethodDescriptor.Marshaller<T> {
         override fun stream(value: T): InputStream {
             // wraps the source in a stream
@@ -58,8 +58,8 @@ public actual fun <Request, Response> methodDescriptor(
 ): MethodDescriptor<Request, Response> {
     return MethodDescriptor.newBuilder<Request, Response>()
         .setFullMethodName(fullMethodName)
-        .setRequestMarshaller(requestCodec.toJvm())
-        .setResponseMarshaller(responseCodec.toJvm())
+        .setRequestMarshaller(requestCodec.toMarshaller())
+        .setResponseMarshaller(responseCodec.toMarshaller())
         .setType(type.asJvm)
         .setSchemaDescriptor(schemaDescriptor)
         .setIdempotent(idempotent)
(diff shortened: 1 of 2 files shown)


grpc/grpc-core/src/jvmMain/kotlin/kotlinx/rpc/grpc/descriptor/MethodDescriptor.jvm.kt
```diff
@@ -31,11 +34,13 @@ internal val MethodType.asJvm: io.grpc.MethodDescriptor.MethodType
 private fun <T> MessageCodec<T>.toJvm(): io.grpc.MethodDescriptor.Marshaller<T> {
```
Mr3zee: Let's rename this one to `toMarshaller`, as this happens only on jvm anyway

grpc/grpc-core/src/nativeMain/kotlin/kotlinx/rpc/grpc/descriptor/MethodDescriptor.native.kt
```diff
@@ -6,13 +6,12 @@ package kotlinx.rpc.grpc.descriptor
...
-import kotlinx.rpc.protobuf.input.stream.InputStream
 
 public actual class MethodDescriptor<Request, Response> internal constructor(
     private val fullMethodName: String,
-    private val requestMarshaller: Marshaller<Request>,
-    private val responseMarshaller: Marshaller<Response>,
-    internal val methodType: MethodType,
+    @InternalRpcApi public val requestCodec: MessageCodec<Request>,
+    @InternalRpcApi public val responseCodec: MessageCodec<Response>,
+    @InternalRpcApi public val methodType: MethodType,
```
Mr3zee: I think it can be public. Previously public were getters, but here we can just go for the properties, can we?

Jozott00: Yeah, I think public should be fine

PR #572, opened 2026-02-12
pb: Add @GeneratedProtoMessage annotation and descriptor object

Jozott00:
**Subsystem**
Protobuf

**Problem Description**
Currently, we store no metadata about a generated protobuf message, such as the message's proto fullName. Additionally, we currently cannot distinguish between a generated proto message and a user-provided one.

This is required for functions that need such metadata information, like the conventional extension functions on the well-known Any type.

**Solution**
This PR adds the `@GeneratedProtoMessage` annotation, which is applied to all generated message interfaces. A FIR diagnostic checker checks if the interface is indeed generated, such that it has a corresponding internal message class and a descriptor object. 

It also adds the `ProtoDescriptor` class, which is internal and currently only contains the message full name. Using the new internal `fun protoDescriptorOf<@GeneratedProtoMessage T>()`, it is possible to retrieve the message's descriptor. The descriptor is generated as part of the message's internal class
```kt
public class MyMessageInternal implements MyMessage {
  public object DESCRIPTOR: ProtoDescriptor<MyMessage> { ... }
}
```

### DIFF
--- a/protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
+++ b/protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
@@ -512,11 +512,11 @@ class ModelToProtobufKotlinCommonGenerator(
             returnType = declaration.name.scoped(),
             comment = Comment.leading(
                 """
-                |Copies the original message, including unknown fields.
-                |```
-                |val copy = original.copy$invocation
-                |```
-            """.trimMargin()
+                    |Copies the original message, including unknown fields.
+                    |```
+                    |val copy = original.copy$invocation
+                    |```
+                """.trimMargin()
             )
         ) {
             code("return this.asInternal().copyInternal(body)".scoped())
@@ -643,10 +643,10 @@ class ModelToProtobufKotlinCommonGenerator(
         val comment = if (declaration.hasPresenceFields) {
             // TODO KRPC-252 protoc-gen: Support type resolution in comments
             Comment.leading(
-            """
-                Interface providing field-presence information for [${declaration.name.fullName()}] messages.
-                Retrieve it via the [${declaration.name.fullName()}.presence] extension property.
-            """.trimIndent()
+                """
+                    Interface providing field-presence information for [${declaration.name.fullName()}] messages.
+                    Retrieve it via the [${declaration.name.fullName()}.presence] extension property.
+                """.trimIndent()
             )
         } else {
             null


compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/ProtoDeclaration.kt
```diff
@@ -0,0 +1,13 @@
...
+
+package kotlinx.rpc.codegen.extension
+
+import org.jetbrains.kotlin.ir.declarations.IrClass
+
+internal class ProtoDeclaration(
+    val message: IrClass,
+    val messageInternal: IrClass,
+    val descriptor: IrClass,
+)
```
Mr3zee: You can move it to the same class as the descriptor, the class is to small for it's own place

compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/ProtoDescriptorGenerator.kt
```diff
@@ -0,0 +1,78 @@
...
+     *
+     * ```
+     * @WithProtoDescriptor(MyMessageInternal.$protoDescriptor::class)
+     * interface MyMessage {}
+     * ```
+     */
+    private fun addObjectAnnotation() {
+        val message = declaration.message
+
+        message.annotations += ctx.vsApi {
```
Mr3zee, line 57: There is actually new API to work with this, but it's in main and not yet here. I can do a rebase, but I'm afraid that with already existing conflicts from `grpc-common` that would be a mess. I can do a rebase after you resolve current conflicts and then you can update this too. Would work?

compiler-plugin/compiler-plugin-k2/src/main/kotlin/kotlinx/rpc/codegen/checkers/FirProtoMessageAnnotationChecker.kt
```diff
@@ -0,0 +1,107 @@
...
+        // and if the internal class has a MyMessageInternal.DESCRIPTOR object
+
+        if (!declaration.isInterface) {
+            reportNonGeneratedMessage()
+        }
+
+        val parentClasses = declaration.parentDeclarationSequence(
+            context.session,
+            null,
+            listOf()
```
Mr3zee: Please add named parameters

compiler-plugin/compiler-plugin-k2/src/main/kotlin/kotlinx/rpc/codegen/checkers/FirProtoMessageAnnotationChecker.kt
```diff
@@ -0,0 +1,107 @@
...
+        ).let {
+            topLevelNames.drop(1).fold(it) { acc, name ->
+                acc.createNestedClassId(name)
+            }
+        }
+
+        val internalDeclaration = context.session.getRegularClassSymbolByClassId(internalClassId)
+            // although this is not safe in general, at the point when the FirRegularClassChecker runs
+            // the symbol should already have been resolved to its FirDeclaration
+            ?.fir
```
Mr3zee: The worst part about that is instability in behaviour and api across versions, not during the resolve. So I suggest to find an alternative here

compiler-plugin/compiler-plugin-k2/src/main/kotlin/kotlinx/rpc/codegen/checkers/FirProtoMessageAnnotationChecker.kt
```diff
@@ -0,0 +1,107 @@
...
+            listOf()
+        )?.filterIsInstance<FirRegularClass>()?.toList()?.reversed() ?: emptyList()
+
+        val topLevelNames = (parentClasses + listOf(declaration))
+            .map { Name.identifier(it.name.asString() + "Internal") }
+
+        // for nested classes, we need to construct ClassId properly using createNestedClassId for each level
+        val internalClassId = ClassId(
+            declaration.symbol.classId.packageFqName,
+            topLevelNames.first()
```
Mr3zee: Named parameters

compiler-plugin/compiler-plugin-k2/src/main/kotlin/kotlinx/rpc/codegen/checkers/FirProtoMessageAnnotationChecker.kt
```diff
@@ -0,0 +1,107 @@
...
+            if (!implementsMessage) {
+                reportNonGeneratedMessage()
+            }
+
+            // check if an internal class has a DESCRIPTOR object declaration
+            val descriptorObject = vsApi {
+                internalDeclaration.declarationsVS(context.session)
+                    .filterIsInstance<FirRegularClassSymbol>()
+                    .find { it.name == ProtoNames.DESCRIPTOR_NAME && it.classKind == ClassKind.OBJECT }
+            }
```
Mr3zee, line 104: Should it also check for the descriptor's super type?

Jozott00: I will add that check

compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/ProtoDeclaration.kt
Jozott00: Do you mean adding the members to the `ProtoDescriptorGenerator`?

Mr3zee: My bad, to the same file, not class

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -101,6 +101,9 @@ class ModelToProtobufKotlinCommonGenerator(
 
         val annotations = mutableListOf<ScopedFormattedString>()
         if (!declaration.isGroup) {
+            annotations.add(
+                "@%T".scoped(FqName.Annotations.GeneratedProtoMessage)
```
Mr3zee: You can use `.scopedAnnotation()`

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -508,11 +512,11 @@ class ModelToProtobufKotlinCommonGenerator(
...
-                    |Copies the original message, including unknown fields.
-                    |```
-                    |val copy = original.copy$invocation
-                    |```
-                """.trimMargin()
+                |Copies the original message, including unknown fields.
+                |```
+                |val copy = original.copy$invocation
+                |```
+            """.trimMargin()
```
Mr3zee: I think the indentation was correct

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -639,10 +643,10 @@ class ModelToProtobufKotlinCommonGenerator(
...
             // TODO KRPC-252 protoc-gen: Support type resolution in comments
             Comment.leading(
-                """
-                    Interface providing field-presence information for [${declaration.name.fullName()}] messages.
-                    Retrieve it via the [${declaration.name.fullName()}.presence] extension property.
-                """.trimIndent()
+            """
+                Interface providing field-presence information for [${declaration.name.fullName()}] messages.
+                Retrieve it via the [${declaration.name.fullName()}.presence] extension property.
+            """.trimIndent()
```
Mr3zee: Same

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -823,6 +827,25 @@ class ModelToProtobufKotlinCommonGenerator(
         }
     }
 
+    private fun CodeGenerator.generateDescriptorObject(declaration: MessageDeclaration) {
+        if (!declaration.isUserFacing) return
+        if (declaration.isGroup) return
+
+        clazz(
+            name = "DESCRIPTOR",
+            annotations = listOf("@%T".scoped(FqName.Annotations.InternalRpcApi)),
```
Mr3zee: `.scopedAnnotation()`

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -823,6 +827,25 @@ class ModelToProtobufKotlinCommonGenerator(
...
+            declarationType = CodeGenerator.DeclarationType.Object,
+            superTypes = listOf("%T<%T>".scoped(FqName.RpcClasses.ProtoDescriptor, declaration.name)),
+        ) {
+            property(
+                name = "fullName",
+                modifiers = "override",
+                type = FqName.Implicits.String.scoped(),
+                value = "\"${declaration.dec.fullName}\"".scoped()
```
Mr3zee, line 844: Just in case, does this account for difference between .proto and kotlin packages?

Jozott00: yes, this comes directly from the Java FieldDescriptor

PR #573, opened 2026-02-13
Protobuf Shortest Unambiguous Names: Implicits

Mr3zee:
**Subsystem**
Protoc-gen Plugin

**ToC**
This is part 1/3 for the feature
 - [x] Add implicit imports
 - [ ] Add Gradle plugin support for platforms
 - [ ] Update codegen to use new scoped strings

**Problem Description**
When we resolve what is the shortest unambiguous name for a type reference in a certain - scope we need to check for known implicitly imported types in the scope. These can be from Kotlin stdlib and from Java stdlib.

**Solution**
Instead of hardcoding the known values, they are scraped from Kdocs and Javadocs of supported versions and stored in the resources.

### DIFF
--- a/update_implicit_types.sh
+++ b/update_implicit_types.sh
@@ -216,6 +216,9 @@ fetch_kotlin_types "kotlin.sequences" "$RESOURCES_DIR/kotlin.sequences.txt"
 echo ""
 fetch_kotlin_types "kotlin.text" "$RESOURCES_DIR/kotlin.text.txt"
 
+echo ""
+fetch_kotlin_types "kotlin.math" "$RESOURCES_DIR/kotlin.math.txt"
+
 # JVM-specific packages
 echo ""
 fetch_kotlin_types "kotlin.jvm" "$RESOURCES_DIR/kotlin.jvm.txt"
--- a/.github/workflows/implicit-imports.yml
+++ b/.github/workflows/implicit-imports.yml
@@ -16,11 +16,11 @@ jobs:
         with:
           fetch-depth: 0
       - name: Update List Of Implicit Imports
-        run: ./update-implicit-types.sh
+        run: ./update_implicit_types.sh
       - name: Check if list is up-to-date
         run: |
           if [[ -n "$(git status --porcelain | grep protoc-gen/common/src/main/resources/implicit-imports/)" ]]; then
-            echo "List of implicit imports is not up to date. Please run './update-implicit-types.sh' and commit changes"
+            echo "List of implicit imports is not up to date. Please run './update_implicit_types.sh' and commit changes"
             git status --porcelain
             git diff --minimal | cat
             exit 1


update_implicit_types.sh
```diff
@@ -0,0 +1,235 @@
...
+    "2.2"
+    "2.3"
+)
+
+# Colors for output
+GREEN='\033[0;32m'
+BLUE='\033[0;34m'
+YELLOW='\033[0;33m'
+RED='\033[0;31m'
+NC='\033[0m' # No Color
```
Jozott00, line 27: fancy fancy ^^

update_implicit_types.sh
Jozott00, line 1: This script's name is currently the only one in kebab-case

update_implicit_types.sh
```diff
@@ -0,0 +1,235 @@
...
+    # The new Dokka format uses: <a ... anchor-label="TypeName" ... data-filterable-set=":kotlin-stdlib...">
+    # We filter for Classlikes (not Functions, Properties, etc.)
+    # Note: Using index.html explicitly to ensure we get the full HTML including deprecated classes
+    # Note: Using -L to follow redirects (latest version redirects index.html to directory)
+    curl -sL "$url" 2>/dev/null | \
+        grep -oE 'anchor-label="[^"]+".* data-filterable-set="[^"]*"' | \
+        grep '%2FClasslikes%2F' | \
+        sed -E 's/anchor-label="([^"]+)".*/\1/' | \
+        grep -E '^[A-Z]' | \
+        LC_ALL=C sort -u > "$output_file"
```
Jozott00, line 99: thought: It might be more stable to scan the stdlib JAR file instead of scanning the published Dokka HTML file, because if the generated Dokka HTML format changes, the script will no longer work.

update_implicit_types.sh
```diff
@@ -0,0 +1,235 @@
...
+fetch_kotlin_types "kotlin.io" "$RESOURCES_DIR/kotlin.io.txt"
+
+echo ""
+fetch_kotlin_types "kotlin.ranges" "$RESOURCES_DIR/kotlin.ranges.txt"
+
+echo ""
+fetch_kotlin_types "kotlin.sequences" "$RESOURCES_DIR/kotlin.sequences.txt"
+
+echo ""
+fetch_kotlin_types "kotlin.text" "$RESOURCES_DIR/kotlin.text.txt"
```
Jozott00, line 217: Not sure if this is actually an issue, but `kotlinx.math` isn't fetched, which is also mentioned as an implicitly imported package in https://kotlinlang.org/spec/packages-and-imports.html

update_implicit_types.sh
Mr3zee: It's fair, I'll see if Claude can manage this

update_implicit_types.sh
Mr3zee: ohhhh, good catch!

update_implicit_types.sh
Mr3zee: 👀

update_implicit_types.sh
Mr3zee: fair

update_implicit_types.sh
Mr3zee: Yeah, that didn't work great
Jar's don't have implicit types, claude is lame with parsing it, also it's not very fast because stdlib is also. Also you need to parse klibs, so yeah, html is better for now. If the format changes - clause can easily update the script and we'll see the failure on CI

PR #574, opened 2026-02-13
Protobuf Shortest Unambiguous Names: Gradle

Mr3zee:
**Subsystem**
Gradle Plugin

**ToC**
This is part 2/3 for the feature
 - [x] Add implicit imports
 - [x] Add Gradle plugin support for platforms
 - [ ] Update codegen to use new scoped strings

**Problem Description**
For properly resolving platform's implicit imports we need to know for which platform the generation is currently in progress.

**Solution**
Detect platforms in build time and pass an option to the protoc-gen plugins


Jozott00 (approved):
lgtm

PR #577, opened 2026-02-19
pb: Add well-known Any extension functions

Jozott00:
**Subsystem**
Protobuf

**Problem Description**
Protobuf libraries provide special functions for the well-known `com.google.protobuf.Any` message type.

**Solution**
This PR adds the following extension functions:

```kotlin
fun <@GeneratedProtoMessage reified T: kotlin.Any> Any.contains(): Boolean
fun <@GeneratedProtoMessage @HasWithCodec reified T : kotlin.Any> Any.Companion.pack(
    value: T,
    urlPrefix: String = "type.googleapis.com",
): Any
fun <@GeneratedProtoMessage @HasWithCodec reified T : kotlin.Any> Any.unpack(): T 
```

which makes it easy to unpack and pack the Any message:
```kotlin
val any = com.google.protobuf.kotlin.Any.pack(Timestamp { seconds = 123 })
if (any.contains<Timestamp>()) {
    val timestamp = any.unpack<Timestamp>()
}
```

### DIFF
--- a/protobuf/protobuf-core/src/commonMain/kotlin/com/google/protobuf/kotlin/anyKotlin.kt
+++ b/protobuf/protobuf-core/src/commonMain/kotlin/com/google/protobuf/kotlin/anyKotlin.kt
@@ -6,12 +6,12 @@ package com.google.protobuf.kotlin
 
 import kotlinx.io.Buffer
 import kotlinx.io.readByteArray
-import kotlinx.rpc.grpc.codec.HasWithCodec
 import kotlinx.rpc.grpc.codec.codec
 import kotlinx.rpc.internal.utils.InternalRpcApi
 import kotlinx.rpc.protobuf.GeneratedProtoMessage
 import kotlinx.rpc.protobuf.internal.protoDescriptorOf
 import kotlin.reflect.KClass
+import kotlin.reflect.KType
 
 /**
  * Checks if this [Any] message contains a message of type [T].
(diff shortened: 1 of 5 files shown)


protobuf/protobuf-core/src/commonMain/kotlin/com/google/protobuf/kotlin/anyKotlin.kt
```diff
@@ -0,0 +1,188 @@
...
+ * }
+ * ```
+ *
+ * @param T the protobuf message type to check for
+ * @param messageClass the [KClass] of the message type to check
+ * @return `true` if this [Any] contains a message of type [messageClass], `false` otherwise
+ * @see pack
+ * @see unpack
+ */
+public fun <@GeneratedProtoMessage T: kotlin.Any> Any.contains(messageClass: KClass<T>): Boolean {
```
Mr3zee, line 59: Please, add one overload with `KType`

protobuf/protobuf-core/src/commonMain/kotlin/com/google/protobuf/kotlin/anyKotlin.kt
```diff
@@ -0,0 +1,188 @@
...
+ * ```
+ *
+ * @param T the protobuf message type to pack
+ * @param value the message instance to pack
+ * @param urlPrefix the URL prefix for the type URL (default: `"type.googleapis.com"`)
+ * @return an [Any] message containing the serialized [value]
+ * @see contains
+ * @see unpack
+ */
+public inline fun <@GeneratedProtoMessage @HasWithCodec reified T : kotlin.Any> Any.Companion.pack(
```
Mr3zee: I think we can annotate `@GeneratedProtoMessage` with `@HasWithCodec` and it should work without both annotations here

PR #579, opened 2026-02-19
cinterop-c: Pin Konan LLVM bundle resolution by Kotlin version and ve…

Jozott00:
**Subsystem**
gRPC/Protobuf/Cinterop

**Problem Description**
Currently, the Konan LLVM bundle locations are hardcoded in the Bazel toolchain configuration.
When updating the project's Kotlin version, this leads to build failures and requires one to manually figure out the new LLVM bundle location and update it at the corresponding toolchain configuration.

**Solution**
This PR adds two new scripts:
- `cinterop-c/toolchain/precompute_konan_llvm_bundles.py`
  Determines the LLVM bundle location for the project's Konan version (optionally also some other version) using the Kotlin Github sources and stores it in the `cinterop-c/toolchain/konan_llvm_bundles.json` file. 
  
- `cinterop-c/toolchain/resolve_konan_llvm_resource_dir.py`
  Reads the `cinterop-c/toolchain/konan_llvm_bundles.json` to find the correct LLVM Clang include directory location based on the KONAN_HOME path. If the location is unknown, it tells the user to update the precomputed bundle locations using `precompute_konan_llvm_bundles.py`

It also adds a new CI workflow that checks if the `konan_llvm_bundles.json` is up to date.

### DIFF
--- a/cinterop-c/toolchain/precompute_konan_llvm_bundles.py
+++ b/cinterop-c/toolchain/precompute_konan_llvm_bundles.py
@@ -47,14 +47,12 @@ def parse_args() -> argparse.Namespace:
     return parser.parse_args()
 
 
-def parse_versions_file(path: Path) -> tuple[str | None, str | None]:
-    """Extract kotlin-lang and kotlin-compiler values from versions-root/libs.versions.toml."""
+def parse_versions_file(path: Path) -> str | None:
+    """Extract kotlin-lang value from versions-root/libs.versions.toml."""
     kotlin_lang: str | None = None
-    kotlin_compiler: str | None = None
     # Keep parsing intentionally narrow to avoid requiring a TOML dependency:
-    # this script only needs two values from versions-root/libs.versions.toml.
+    # this script only needs kotlin-lang from versions-root/libs.versions.toml.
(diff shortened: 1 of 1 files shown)


cinterop-c/toolchain/precompute_konan_llvm_bundles.py
```diff
@@ -0,0 +1,173 @@
...
+            continue
+
+        kotlin_compiler_match = kotlin_compiler_re.match(line)
+        if kotlin_compiler_match:
+            kotlin_compiler = kotlin_compiler_match.group(1)
+
+    return kotlin_lang, kotlin_compiler
+
+
+def resolve_default_kotlin_compiler_version(repo_root: Path) -> str:
```
Mr3zee, line 68: I don't think it should use compiler version, but rather lang version, should it?

Jozott00: I believe the shipped LLVM bundle is tightly coupled to the compiler version that is used for compilation.

Mr3zee: Yeah, but this version in our configs is only used for compiler plugin compilation and not for the project compilation

Jozott00: That's true 🤔 I will change it

PR #581, opened 2026-02-24
Update api for kotlin-master (2.4.0)

Mr3zee:
**Subsystem**
Compiler Plugin

**Problem Description**
Kotlin master (and 261 IDE Kotlin) are not compatible with our grpc version of the plugin

**Solution**
MAKE COMPATIBLE!

### DIFF
--- a/compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcIrContext.kt
+++ b/compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcIrContext.kt
@@ -7,11 +7,13 @@ package kotlinx.rpc.codegen.extension
 import kotlinx.rpc.codegen.VersionSpecificApi
 import org.jetbrains.kotlin.backend.common.extensions.IrPluginContext
 import org.jetbrains.kotlin.ir.declarations.IrEnumEntry
+import org.jetbrains.kotlin.ir.declarations.IrFile
 import org.jetbrains.kotlin.ir.declarations.IrSimpleFunction
 import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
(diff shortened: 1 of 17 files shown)


compiler-plugin/compiler-plugin-backend/src/main/kotlin/kotlinx/rpc/codegen/extension/RpcIrContext.kt
```diff
@@ -22,6 +24,12 @@ import org.jetbrains.kotlin.platform.jvm.isJvm
 import org.jetbrains.kotlin.platform.konan.isNative
 import org.jetbrains.kotlin.types.Variance
 
+/**
+ * IMPORTANT!!!
+ * Difference between a built-in reference and a regular one:
+ * - Built-in - lib classes (kotlinx.rpc.*) and stdlib
+ * - Regular - user classes (kotlinx.serialization.*, kotlinx.coroutines.*) from other libs
+ */
```
Mr3zee, line 32: @Jozott00 FYI

compiler-plugin/compiler-plugin-backend/src/main/templates/kotlinx/rpc/codegen/VersionSpecificApiImpl.kt
Jozott00, line 288: question: Why are kotlinx.rpc classes resolvable using the `finderForBuiltins` function? Is this because the compiler plugin registers the kotlinx.rpc modules as built-in?

Mr3zee: I was also confused, but I think this is naming problem basically.
This is all IC stuff, the reference is here:
https://github.com/JetBrains/kotlin/blob/e5b98e9f15e20ae415ae9db034cc296818031586/compiler/ir/backend.common/src/org/jetbrains/kotlin/backend/common/extensions/IrPluginContext.kt#L74

PR #582, opened 2026-02-24
Update platform tables fro gRPC branch

Mr3zee:
**Subsystem**
Docs

I'm sick of the failing actions 😄

### DIFF
--- a/gradle-conventions/src/main/kotlin/util/targets/kmpConfig.kt
+++ b/gradle-conventions/src/main/kotlin/util/targets/kmpConfig.kt
@@ -14,6 +14,7 @@ import org.jetbrains.kotlin.konan.target.HostManager
 import util.kotlinVersionParsed
 import util.other.optionalProperty
 import util.other.optionalPropertyValue
+import util.tasks.DumpPlatformsTask
 import java.io.File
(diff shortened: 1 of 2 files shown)


Mr3zee:
@Jozott00 Well I fixed it 😅 thanks for the hint!

gradle-conventions/src/main/kotlin/util/targets/kmpConfig.kt
```diff
@@ -100,7 +101,7 @@ class KmpConfig(
                     lookupTable = nativeLookup,
                 ) &&
                 !optionalPropertyValue(targetFunction.name, "exclude") &&
-                !(isAppleOnLinuxByRequirement && isApple)
+                !(isAppleOnLinuxByRequirement && isApple && !gradle.startParameter.taskNames.contains(DumpPlatformsTask.NAME))
```
Jozott00, line 108: Maybe it wouldn't be a bad idea to write a comment about it. I'm sure I'll be confused when I read it in a month xD

Mr3zee: truee

PR #583, opened 2026-02-25
Grpc/merge annotations

Mr3zee:
**Subsystem**
Compiler Plugin

**Problem Description**
We have two annotations generated for generated proto messages, `@GeneratedProtoMessage` and `WithCodec`. 

**Solution**
One is enough

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 [![Kotlin Experimental](https://kotl.in/badges/experimental.svg)](https://kotlinlang.org/docs/components-stability.html)
 [![Official JetBrains project](http://jb.gg/badges/official.svg)](https://confluence.jetbrains.com/display/ALL/JetBrains+on+GitHub)
-[![Kotlin](https://img.shields.io/badge/kotlin-2.0.0--2.3.0-blue.svg?logo=kotlin)](http://kotlinlang.org)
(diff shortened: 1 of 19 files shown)


core/src/jvmMain/kotlin/kotlinx/rpc/internal/internalServiceDescriptorOf.jvm.kt
```diff
@@ -5,20 +5,18 @@
...
-    return try {
-        return kClass.java.classLoader
-            .loadClass(className)
-            ?.kotlin
-            ?.companionObjectInstance
-    } catch (_: ClassNotFoundException) {
-        null
-    }
+    val codecClass = kClass.findAnnotation<WithServiceDescriptor>()?.stub ?: return null
+    return codecClass.objectInstance
```
Jozott00, line 21: `codecClass` copy paste leftover

Jozott00 (approved):
Nice!

core/src/jvmMain/kotlin/kotlinx/rpc/internal/internalServiceDescriptorOf.jvm.kt
Mr3zee: oops, thanks!
I'll fix it in codecs rename pr

PR #586, opened 2026-02-25
Moved companion object generation into the compiler plugin

Mr3zee:
**Subsystem**
Compiler Plugin, Protoc Plugin

**Problem Description**
Protoc plugin generated companion objects in source for all messages

**Solution**
Move the generation into the compiler plugin and enable IDE support with KEFS!!!


Jozott00 (approved):
Truly magical ^^

sadikukumukiza-netizen:
Elamu

PR #587, opened 2026-02-25
pb: Add support for extensions

Jozott00:
**Subsystem**
Protobuf

**Problem Description**
Protobuf allows the use of extensions, which are currently not supported.

**Solution**
Add extension descriptors that describe the properties of an dynamic extension field including decoding and encoding. When a proto message defines such extension fields, protoc generates a construction call.

To ease the access of extensions, we generate extension getters and setters on the extendee message class.

```kotlin
val ExtensionBase.int32: Int? get() = asInternal().getExtensionValue(ExtensionExtendKtExtensions.int32)

val ExtensionBase.Companion.int32: ProtoExtensionDescriptor<ExtensionBase, Int> get() = ExtensionExtendKtExtensions.int32

val ExtensionBasePresence.hasInt32: Boolean get() = (this as InternalPresenceObject).hasExtension(ExtensionExtendKtExtensions.int32)

var ExtensionBase.Builder.int32: Int?
    get() = asInternal().getExtensionValue(ExtensionExtendKtExtensions.int32)
    set(value) { asInternal().setExtensionValue(ExtensionExtendKtExtensions.int32, value) }
```

While encoding doesn’t require any registration of the extension descriptor, users must register them when they intend to decode them. This is done by building a `ProtoExtensionRegistry` and adding it to the `ProtoConfig` that is passed during decoding.
```kotlin
val registry = buildProtoExtensionRegistry {
   +ExtensionBase.int32 // or register(ExtensionBase.int32)
}
val config = ProtoConfig(extensionRegistry=registry)
// ...
```

### DIFF
--- a/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf_test_messages/edition2023/_rpc_internal/TestMessagesEdition2023.kt
+++ b/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf_test_messages/edition2023/_rpc_internal/TestMessagesEdition2023.kt
@@ -1178,8 +1178,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: Int by MsgFieldDelegate(PresenceIndices.value) { 0 }
 
-        private val _owner: MapInt32Int32EntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -1250,8 +1248,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
 
         var key: Long by MsgFieldDelegate(PresenceIndices.key) { 0L }
         var value: Long by MsgFieldDelegate(PresenceIndices.value) { 0L }
-
-        private val _owner: MapInt64Int64EntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -1324,8 +1320,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
         var key: UInt by MsgFieldDelegate(PresenceIndices.key) { 0u }
         var value: UInt by MsgFieldDelegate(PresenceIndices.value) { 0u }
 
-        private val _owner: MapUint32Uint32EntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -1396,8 +1390,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
 
         var key: ULong by MsgFieldDelegate(PresenceIndices.key) { 0uL }
         var value: ULong by MsgFieldDelegate(PresenceIndices.value) { 0uL }
-
-        private val _owner: MapUint64Uint64EntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -1470,8 +1462,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: Int by MsgFieldDelegate(PresenceIndices.value) { 0 }
 
-        private val _owner: MapSint32Sint32EntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -1542,8 +1532,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
 
         var key: Long by MsgFieldDelegate(PresenceIndices.key) { 0L }
         var value: Long by MsgFieldDelegate(PresenceIndices.value) { 0L }
-
-        private val _owner: MapSint64Sint64EntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -1616,8 +1604,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
         var key: UInt by MsgFieldDelegate(PresenceIndices.key) { 0u }
         var value: UInt by MsgFieldDelegate(PresenceIndices.value) { 0u }
 
-        private val _owner: MapFixed32Fixed32EntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -1688,8 +1674,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
 
         var key: ULong by MsgFieldDelegate(PresenceIndices.key) { 0uL }
         var value: ULong by MsgFieldDelegate(PresenceIndices.value) { 0uL }
-
-        private val _owner: MapFixed64Fixed64EntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -1762,8 +1746,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: Int by MsgFieldDelegate(PresenceIndices.value) { 0 }
 
-        private val _owner: MapSfixed32Sfixed32EntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -1834,8 +1816,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
 
         var key: Long by MsgFieldDelegate(PresenceIndices.key) { 0L }
         var value: Long by MsgFieldDelegate(PresenceIndices.value) { 0L }
-
-        private val _owner: MapSfixed64Sfixed64EntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -1908,8 +1888,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: Float by MsgFieldDelegate(PresenceIndices.value) { 0.0f }
 
-        private val _owner: MapInt32FloatEntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -1980,8 +1958,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
 
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: Double by MsgFieldDelegate(PresenceIndices.value) { 0.0 }
-
-        private val _owner: MapInt32DoubleEntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -2054,8 +2030,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
         var key: Boolean by MsgFieldDelegate(PresenceIndices.key) { false }
         var value: Boolean by MsgFieldDelegate(PresenceIndices.value) { false }
 
-        private val _owner: MapBoolBoolEntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -2126,8 +2100,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
 
         var key: String by MsgFieldDelegate(PresenceIndices.key) { "" }
         var value: String by MsgFieldDelegate(PresenceIndices.value) { "" }
-
-        private val _owner: MapStringStringEntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -2200,8 +2172,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
         var key: String by MsgFieldDelegate(PresenceIndices.key) { "" }
         var value: ByteArray by MsgFieldDelegate(PresenceIndices.value) { byteArrayOf() }
 
-        private val _owner: MapStringBytesEntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -2272,8 +2242,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
 
         var key: String by MsgFieldDelegate(PresenceIndices.key) { "" }
         var value: TestAllTypesEdition2023.NestedMessage by MsgFieldDelegate(PresenceIndices.value) { NestedMessageInternal() }
-
-        private val _owner: MapStringNestedMessageEntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -2346,8 +2314,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
         var key: String by MsgFieldDelegate(PresenceIndices.key) { "" }
         var value: ForeignMessageEdition2023 by MsgFieldDelegate(PresenceIndices.value) { ForeignMessageEdition2023Internal() }
 
-        private val _owner: MapStringForeignMessageEntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -2419,8 +2385,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
         var key: String by MsgFieldDelegate(PresenceIndices.key) { "" }
         var value: TestAllTypesEdition2023.NestedEnum by MsgFieldDelegate(PresenceIndices.value) { TestAllTypesEdition2023.NestedEnum.FOO }
 
-        private val _owner: MapStringNestedEnumEntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -2491,8 +2455,6 @@ class TestAllTypesEdition2023Internal: TestAllTypesEdition2023.Builder, Internal
 
         var key: String by MsgFieldDelegate(PresenceIndices.key) { "" }
         var value: ForeignEnumEdition2023 by MsgFieldDelegate(PresenceIndices.value) { ForeignEnumEdition2023.FOREIGN_FOO }
-
-        private val _owner: MapStringForeignEnumEntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -6440,15 +6402,14 @@ fun TestAllTypesEdition2023.NestedEnum.Companion.fromNumber(number: Int): TestAl
 
 @InternalRpcApi
 object TestMessagesEdition2023KtExtensions {
-    val extensionInt32: InternalExtensionDescriptor<TestAllTypesEdition2023,  Int> =
-        InternalExtensionDescriptor.int32(
+
+    val extensionInt32: InternalExtensionDescriptor<TestAllTypesEdition2023,  Int> =         InternalExtensionDescriptor.int32(
             fieldNumber = 120,
             name = "extensionInt32",
             extendee = TestAllTypesEdition2023::class,
         )
 
-    val groupliketype: InternalExtensionDescriptor<TestAllTypesEdition2023,  GroupLikeType> =
-        InternalExtensionDescriptor.message(
+    val groupliketype: InternalExtensionDescriptor<TestAllTypesEdition2023,  GroupLikeType> =         InternalExtensionDescriptor.message(
             fieldNumber = 121,
             name = "groupliketype",
             extendee = TestAllTypesEdition2023::class,
@@ -6459,8 +6420,7 @@ object TestMessagesEdition2023KtExtensions {
             decodeWith = { value, decoder, config -> GroupLikeTypeInternal.decodeWith(value.asInternal(), decoder, config) },
         )
 
-    val delimitedExt: InternalExtensionDescriptor<TestAllTypesEdition2023,  GroupLikeType> =
-        InternalExtensionDescriptor.message(
+    val delimitedExt: InternalExtensionDescriptor<TestAllTypesEdition2023,  GroupLikeType> =         InternalExtensionDescriptor.message(
             fieldNumber = 122,
             name = "delimitedExt",
             extendee = TestAllTypesEdition2023::class,
--- a/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf/conformance/_rpc_internal/Conformance.kt
+++ b/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf/conformance/_rpc_internal/Conformance.kt
@@ -41,8 +41,6 @@ class TestStatusInternal: TestStatus.Builder, InternalMessage(fieldsWithPresence
     override var failureMessage: String by MsgFieldDelegate { "" }
     override var matchedName: String by MsgFieldDelegate { "" }
 
-    private val _owner: TestStatusInternal = this
-
     override fun hashCode(): Int {
         checkRequiredFields()
         var result = name.hashCode()
@@ -143,8 +141,6 @@ class FailureSetInternal: FailureSet.Builder, InternalMessage(fieldsWithPresence
     internal var _unknownFieldsEncoder: WireEncoder? = null
 
     override var test: List<TestStatus> by MsgFieldDelegate { mutableListOf() }
-
-    private val _owner: FailureSetInternal = this
 
     override fun hashCode(): Int {
         checkRequiredFields()
@@ -418,8 +414,6 @@ class ConformanceResponseInternal: ConformanceResponse.Builder, InternalMessage(
 
     override var result: ConformanceResponse.Result? = null
 
-    private val _owner: ConformanceResponseInternal = this
-
     override fun hashCode(): Int {
         checkRequiredFields()
         var result = (result?.oneOfHashCode() ?: 0)
@@ -576,8 +570,6 @@ class JspbEncodingConfigInternal: JspbEncodingConfig.Builder, InternalMessage(fi
     internal var _unknownFieldsEncoder: WireEncoder? = null
 
     override var useJspbArrayAnyFormat: Boolean by MsgFieldDelegate { false }
-
-    private val _owner: JspbEncodingConfigInternal = this
 
     override fun hashCode(): Int {
         checkRequiredFields()
--- a/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf_test_messages/editions/proto2/_rpc_internal/TestMessagesProto2Editions.kt
+++ b/tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf_test_messages/editions/proto2/_rpc_internal/TestMessagesProto2Editions.kt
@@ -1621,8 +1621,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: Int by MsgFieldDelegate(PresenceIndices.value) { 0 }
 
-        private val _owner: MapInt32Int32EntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -1693,8 +1691,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
 
         var key: Long by MsgFieldDelegate(PresenceIndices.key) { 0L }
         var value: Long by MsgFieldDelegate(PresenceIndices.value) { 0L }
-
-        private val _owner: MapInt64Int64EntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -1767,8 +1763,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
         var key: UInt by MsgFieldDelegate(PresenceIndices.key) { 0u }
         var value: UInt by MsgFieldDelegate(PresenceIndices.value) { 0u }
 
-        private val _owner: MapUint32Uint32EntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -1839,8 +1833,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
 
         var key: ULong by MsgFieldDelegate(PresenceIndices.key) { 0uL }
         var value: ULong by MsgFieldDelegate(PresenceIndices.value) { 0uL }
-
-        private val _owner: MapUint64Uint64EntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -1913,8 +1905,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: Int by MsgFieldDelegate(PresenceIndices.value) { 0 }
 
-        private val _owner: MapSint32Sint32EntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -1985,8 +1975,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
 
         var key: Long by MsgFieldDelegate(PresenceIndices.key) { 0L }
         var value: Long by MsgFieldDelegate(PresenceIndices.value) { 0L }
-
-        private val _owner: MapSint64Sint64EntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -2059,8 +2047,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
         var key: UInt by MsgFieldDelegate(PresenceIndices.key) { 0u }
         var value: UInt by MsgFieldDelegate(PresenceIndices.value) { 0u }
 
-        private val _owner: MapFixed32Fixed32EntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -2131,8 +2117,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
 
         var key: ULong by MsgFieldDelegate(PresenceIndices.key) { 0uL }
         var value: ULong by MsgFieldDelegate(PresenceIndices.value) { 0uL }
-
-        private val _owner: MapFixed64Fixed64EntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -2205,8 +2189,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: Int by MsgFieldDelegate(PresenceIndices.value) { 0 }
 
-        private val _owner: MapSfixed32Sfixed32EntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -2277,8 +2259,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
 
         var key: Long by MsgFieldDelegate(PresenceIndices.key) { 0L }
         var value: Long by MsgFieldDelegate(PresenceIndices.value) { 0L }
-
-        private val _owner: MapSfixed64Sfixed64EntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -2351,8 +2331,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: Boolean by MsgFieldDelegate(PresenceIndices.value) { false }
 
-        private val _owner: MapInt32BoolEntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -2423,8 +2401,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
 
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: Float by MsgFieldDelegate(PresenceIndices.value) { 0.0f }
-
-        private val _owner: MapInt32FloatEntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -2497,8 +2473,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: Double by MsgFieldDelegate(PresenceIndices.value) { 0.0 }
 
-        private val _owner: MapInt32DoubleEntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
@@ -2569,8 +2543,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
 
         var key: Int by MsgFieldDelegate(PresenceIndices.key) { 0 }
         var value: TestAllTypesProto2.NestedMessage by MsgFieldDelegate(PresenceIndices.value) { NestedMessageInternal() }
-
-        private val _owner: MapInt32NestedMessageEntryInternal = this
 
         override fun hashCode(): Int {
             checkRequiredFields()
@@ -2643,8 +2615,6 @@ class TestAllTypesProto2Internal: TestAllTypesProto2.Builder, InternalMessage(fi
         var key: Boolean by MsgFieldDelegate(PresenceIndices.key) { false }
         var value: Boolean by MsgFieldDelegate(PresenceIndices.value) { false }
 
-        private val _owner: MapBoolBoolEntryInternal = this
-
         override fun hashCode(): Int {
             checkRequiredFields()
             var result = if (presenceMask[0]) key.hashCode() else 0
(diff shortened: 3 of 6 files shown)


Jozott00:
@Mr3zee Note: I will add the missing unit tests in a follow up PR. I have to fix some issue with unknown fields first.

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalExtensionDescriptor.kt
```diff
@@ -0,0 +1,602 @@
...
+    // Human-readable/debug name (used in toString rendering of extension values).
+    public val name: String,
+    // Canonical wire type for this extension in non-packed form.
+    public val wireType: WireType,
+    // Accepts alternative wire representations (e.g., packed/unpacked compatibility).
+    public val acceptedWireTypes: Set<WireType> = setOf(wireType),
+    // Default value exposed by generated extension getters when extension is unset.
+    public val defaultValue: Lazy<V>,
+    // Computes encoded size for value currently stored in the extension map.
+    public val size: (Int, Any) -> Int,
```
Mr3zee: Please add args desc

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalExtensionDescriptor.kt
```diff
@@ -0,0 +1,602 @@
...
+    // Needed for runtime casts when values are stored in erased Any slots.
+    public val valueType: KClass<V>,
+    // Protobuf field number used for extension lookup and wire encoding.
+    public override val fieldNumber: Int,
+    // Human-readable/debug name (used in toString rendering of extension values).
+    public val name: String,
+    // Canonical wire type for this extension in non-packed form.
+    public val wireType: WireType,
+    // Accepts alternative wire representations (e.g., packed/unpacked compatibility).
+    public val acceptedWireTypes: Set<WireType> = setOf(wireType),
```
Mr3zee: I guess not alternative, but just all

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalExtensionDescriptor.kt
```diff
@@ -0,0 +1,602 @@
...
+    // Default value exposed by generated extension getters when extension is unset.
+    public val defaultValue: Lazy<V>,
+    // Computes encoded size for value currently stored in the extension map.
+    public val size: (Int, Any) -> Int,
+    // Encodes value currently stored in the extension map.
+    public val encode: (WireEncoder, Int, Any, ProtobufConfig?) -> Unit,
```
Mr3zee: Same about args, also all further functions

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalExtensionDescriptor.kt
```diff
@@ -0,0 +1,602 @@
...
+    // Helper used by packedRepeated(...) to emit packed payloads for element descriptors.
+    private val encodePacked: ((WireEncoder, Int, List<V>) -> Unit)? = null,
+    // Helper used by packedRepeated(...) to decode packed element payloads.
+    private val decodePackedValues: ((WireDecoder) -> List<V>)? = null,
+): ProtoExtensionDescriptor<T, V> {
+    // Packed-ness is fully determined by whether a packed decoder exists.
+    public val isPacked: Boolean get() = decodePacked != null
+
+    public companion object {
+
```
Mr3zee: NL not needed

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalExtensionDescriptor.kt
```diff
@@ -0,0 +1,602 @@
...
+import kotlinx.rpc.internal.utils.InternalRpcApi
+import kotlinx.rpc.protobuf.ProtoExtensionDescriptor
+import kotlinx.rpc.protobuf.ProtobufConfig
+import kotlin.reflect.KClass
+import kotlin.reflect.cast
+import kotlin.reflect.safeCast
+
+@InternalRpcApi
+public class InternalExtensionDescriptor<@GeneratedProtoMessage T : Any, V : Any> private constructor(
```
Mr3zee, line 18: All of the values in the class look like generic generators. Can we reuse them in codegen for regular fields?

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -52,29 +52,42 @@ class ModelToProtobufKotlinCommonGenerator(
     )
```
Mr3zee, line 52: please add `_extensions` here

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalMessage.kt
```diff
@@ -8,16 +8,79 @@ import kotlinx.io.Buffer
...
+    public fun extensionsSize(): Int {
+        var result = 0
+        _extensions.forEach { (key, value) ->
+            result += value.descriptor.size(key, value.value)
+        }
+        return result
+    }
+
+    protected fun copyExtensionsFrom(other: InternalMessage) {
+        // we don't have to copy the value's themself, as they
```
Mr3zee: cliffhanger 😅

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/ProtoExtension.kt
```diff
@@ -0,0 +1,124 @@
...
+package kotlinx.rpc.protobuf
+
+import kotlinx.rpc.protobuf.internal.GeneratedProtoMessage
+import kotlin.reflect.KClass
+
+/**
+ * Protobuf extension descriptor.
+ *
+ * This interface is not meant to be implemented by users.
+ * But is used by the protoc compiler plugin to generate extension descriptors.
```
Mr3zee: `protoc compiler plugin` is two things in one. Either `protoc plugin` or `compiler plugin` or both

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/ProtoExtension.kt
```diff
@@ -0,0 +1,124 @@
...
+ * Creates a [ProtoExtensionRegistry] using a DSL-style builder block.
+ *
+ * Example:
+ * ```
+ * val registry = buildProtoExtensionRegistry {
+ *     +MyMessage.myExtension
+ * }
+ * ```
+ */
+public fun buildProtoExtensionRegistry(builder: ProtoExtensionRegistry.Builder.() -> Unit): ProtoExtensionRegistry {
```
Mr3zee: Can be just `ProtoExtensionRegistry` as a constructor function

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/ProtoExtension.kt
```diff
@@ -0,0 +1,124 @@
...
+    return ProtoExtensionRegistry().also {
+        it.newBuilder().apply(builder)
+    }
+}
+
+/**
+ * Runtime registry of known extension descriptors.
+ *
+ * The registry can be attached to a [ProtobufConfig] to enable extension parsing.
```
Mr3zee: `attached` is rather vague. `Passed to`?

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/ProtoExtension.kt
```diff
@@ -0,0 +1,124 @@
...
+ * Example:
+ * ```
+ * val registry = buildProtoExtensionRegistry {
+ *     +MyFileExtensions.myExtension
+ *     +MyFileExtensions.anotherExtension
+ * }
+ * val config = ProtobufConfig(extensionRegistry = registry)
+ * ```
+ */
+public class ProtoExtensionRegistry internal constructor() {
```
Mr3zee, line 59: I think is other places with use full `Protobuf` suffix for names

protobuf/protobuf-api/src/commonTest/kotlin/kotlinx/rpc/protobuf/test/ProtobufSizeCalculationTest.kt
```diff
@@ -8,10 +8,16 @@ package kotlinx.rpc.protobuf.test
...
+import kotlinx.rpc.protobuf.internal.InternalExtensionDescriptor
 import kotlin.test.Test
 import kotlin.test.assertEquals
 
 class ProtobufSizeCalculationTest {
+    private val stringExtension = InternalExtensionDescriptor.string(
+        fieldNumber = 99,
+        name = "testString",
+        extendee = ExtensionBase::class,
+    )
```
Mr3zee, line 21: Love the definition!

protobuf/protobuf-api/src/commonTest/kotlin/kotlinx/rpc/protobuf/test/ProtobufSizeCalculationTest.kt
```diff
@@ -71,4 +77,146 @@ class ProtobufSizeCalculationTest {
...
+            label = "nested message extension",
+        )
+    }
+
+    @Test
+    fun testLengthDelimitedScalarExtensionSizeIsCorrect() {
+        val message = ExtensionBaseInternal().apply {
+            setExtensionValue(stringExtension, "hello")
+        }
+
```
Mr3zee, line 131: Is this an intended api for using it?

protobuf/protobuf-api/src/commonTest/kotlin/kotlinx/rpc/protobuf/test/ProtobufSizeCalculationTest.kt
```diff
@@ -71,4 +77,146 @@ class ProtobufSizeCalculationTest {
...
+        assertEquals(
+            actualSize,
+            declaredSize,
+            "The declared _size ($declaredSize) should match the actual encoded size ($actualSize) for $label. Actual bytes: ${bytes.toHexString()}"
+        )
+    }
+
+    // TODO: Add extension size tests for map-typed extensions once extension descriptor generation supports FieldType.Map.
+    // TODO: Add extension size tests for oneof-typed extensions if/when extension descriptor generation supports FieldType.OneOf.
```
Mr3zee: Please add tickets or tests if the support is done

protobuf/protobuf-api/src/commonTest/kotlin/kotlinx/rpc/protobuf/test/ProtoExtensionTest.kt
```diff
@@ -0,0 +1,342 @@
...
+    }
+
+    @Test
+    fun `test extension registry - register single extension`() {
+        val registry = buildProtoExtensionRegistry {
+            register(ExtensionBase.int32)
+        }
+
+        val extensions = registry.allExtensionsForMessage(ExtensionBase::class)
+        assertEquals(1, extensions.size)
```
Mr3zee: Registring `ExtensionBase.int32` twice

protobuf/protobuf-wkt/src/commonMain/generated-code/kotlin-multiplatform/com/google/protobuf/kotlin/_rpc_internal/Any.kt
```diff
@@ -36,6 +37,8 @@ public class AnyInternal: Any.Builder, InternalMessage(fieldsWithPresence = 0) {
     public override var typeUrl: String by MsgFieldDelegate { "" }
     public override var value: ByteArray by MsgFieldDelegate { byteArrayOf() }
 
+    private val _owner: AnyInternal = this
```
Mr3zee: Don't see a usage for it. Also can conflict with user names unless added to the blacklist in the generator

protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/model/model.kt
```diff
@@ -83,6 +89,15 @@ data class MessageDeclaration(
                         !(it.dec.defaultValue as ByteString).isEmpty
             }
     }
+
+    val extensionRanges: List<IntRange> by lazy {
+        dec.toProto().extensionRangeList.map { it.start..it.end }
+    }
+
+    val mayHaveExtensions: Boolean by lazy {
```
Mr3zee: Why may?

protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/CodeGenerator.kt
```diff
@@ -235,6 +235,72 @@ open class CodeGenerator(
...
+            code(function.wrapIn { "$it()" })
+            return
+        }
+
+        selectNames {
+            addLine(function.wrapIn { "$it(" })
+        }
+
+        withNextIndent {
+            for ((name, value) in namedArgs) {
```
Mr3zee, line 252: You can do a trick with counting the message length before deciding whether you want to put args on an NL or not, e.g. to have `foo(a = a)` on one line if it is not long

protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/CodeGenerator.kt
```diff
@@ -267,13 +338,75 @@ open class CodeGenerator(
...
+                    newLine()
+                    newLine()
+                }
+            }
+        }
+
+        this.needsNewLineAfterDeclaration = needsNewLineAfterDeclaration
+    }
+
+    fun complexProperty(
```
Mr3zee: `complex` is a bit cryptic, maybe just `property` is good too?

protoc-gen/common/src/test/kotlin/kotlinx/rpc/protoc/gen/test/CodeGeneratorTest.kt
```diff
@@ -116,6 +116,55 @@ class CodeGeneratorTest {
...
+                }
+            }
+        }
+
+        assertEquals(
+            """
+            val value: String =
+                buildString {
+                    append("hello")
+                }
```
Mr3zee: I would expect 
```
val value: Strin = buildString {
    append("hello")
}
```

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -2012,12 +2201,226 @@ class ModelToProtobufKotlinCommonGenerator(
...
+    extensions.isNotEmpty() || nestedDeclarations.any { it.hasExtensionsRecursively() }
+
+private fun FileDeclaration.allExtensions(): List<FieldDeclaration> =
+    extensions + allNestedExtensions()
+
+private fun FileDeclaration.allNestedExtensions(): List<FieldDeclaration> =
+    (messageDeclarations + messageDeclarations.flatMap(MessageDeclaration::allNestedRecursively))
+        .flatMap { it.extensions }
+
+private fun FileDeclaration.messageDeclarationsWithExtensionGroups(): List<MessageDeclaration> {
```
Mr3zee: This is called quite a few times, maybe cache it in `FileDeclaration` class?

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -250,6 +264,19 @@ class ModelToProtobufKotlinCommonGenerator(
                 )
             }
 
+            // this property is required to implement the InternalPresenceObject interface
+            // for every possible case: if we have a nested message type with the same name
+            // like MyClass1Internal.StringInternal.StringInternal, we cannot use "this@StringInternal"
+            // as it is ambiguous.
+            // to workaround this, we add the "_owner" property to the internal message, that can be
+            // used by the presence object to reference its message.
```
Mr3zee: good! You can add a check that the perense object is generated at all

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -651,6 +694,20 @@ class ModelToProtobufKotlinCommonGenerator(
...
+        // InternalMessage
+        function(
+            name = "copyInternal",
+            modifiers = "override",
+            returnType = declaration.internalClassName.scoped(),
+        ) {
+            if (declaration.isUserFacing) {
+                code("return copyInternal { }".scoped())
+            } else {
+                code("return this".scoped())
```
Mr3zee, line 698: Why are only user-facing actually copied?

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -818,16 +880,111 @@ class ModelToProtobufKotlinCommonGenerator(
...
 
+    private fun CodeGenerator.generateProtoExtensionProperty(declaration: FieldDeclaration) {
+        val name = declaration.name
+        val extendee = declaration.containingType.value
+        val descriptorRef = "%T".scoped(requireNotNull(declaration.extensionDescriptorName) {
+            "Missing extension descriptor name for ${declaration.name}"
+        })
+
+        var value = "asInternal().getExtensionValue(".scoped()
+            .merge(descriptorRef) { prefix, descriptorRef -> "$prefix$descriptorRef)" }
```
Mr3zee: Can be simplified to `"asInternal().getExtensionValue(%T)".scoped(descriptorRef)`

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -818,16 +880,111 @@ class ModelToProtobufKotlinCommonGenerator(
...
+        if (!declaration.nullable) {
+            // if the field is non-nullable, e.g. a message, we fallback to the defined default value
+            val default = descriptorRef.wrapIn { descriptorRef -> "$descriptorRef.defaultValue.value" }
+            value = default
+                .merge(value) { default, value ->
+                    "$value ?: $default"
+                }
```
Mr3zee: Wrong order of merge parameters

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -818,16 +880,111 @@ class ModelToProtobufKotlinCommonGenerator(
...
+        }
+
+        // val MyMessage.myExtensionField: FieldType? get()
+        //  = asInternal().getExtensionValue(MyProtoFileKtExtensions.myExtensionField)
```
Mr3zee: Equals sign usually  stays on the first line

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -818,16 +880,111 @@ class ModelToProtobufKotlinCommonGenerator(
...
+        //  = MyProtoFileKtExtensions.myExtensionField
+        property(
+            name = name,
+            contextReceiver = "%T.Companion".scoped(extendee.name),
+            type = "%T<%T".scoped(
+                FqName.RpcClasses.ProtoExtensionDescriptor,
+                extendee.name,
+            ).merge(declaration.typeFqNameNonNullable()) { first, second ->
+                "$first, $second>"
+            },
```
Mr3zee, line 911: Can be simplified to `"%T<%T, %T>".scoped(...)`

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -818,16 +880,111 @@ class ModelToProtobufKotlinCommonGenerator(
...
+
+        // val MyMessage.hasMyExtensionField: Boolean get()
+        //  = (this as InternalPresenceObject).hasExtension(MyProtoFileKtExtensions.int32)
+        property(
+            name = "has" + name.capitalize(),
+            contextReceiver = extendee.presenceInterfaceName.scoped(),
+            type = FqName.Implicits.Boolean.scoped(),
+            propertyInitializer = CodeGenerator.PropertyInitializer.GETTER,
+            value = "(this as %T).hasExtension(".scoped(FqName.RpcClasses.InternalPresenceObject)
+                .merge(descriptorRef) { prefix, descriptorRef -> "$prefix$descriptorRef)" }
```
Mr3zee: Same here

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -818,16 +880,111 @@ class ModelToProtobufKotlinCommonGenerator(
...
+
+        property(
+            name = name,
+            contextReceiver = extendee.builderClassName.scoped(),
+            type = declaration.typeFqName(),
+            propertyInitializer = CodeGenerator.PropertyInitializer.GETTER,
+            isVar = true,
+            value = value,
+            setter = "asInternal().setExtensionValue(".scoped()
+                .merge(descriptorRef) { prefix, descriptorRef -> "$prefix$descriptorRef, value)" },
```
Mr3zee: and here

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -2012,12 +2201,226 @@ class ModelToProtobufKotlinCommonGenerator(
...
+            "Missing extension descriptor name for ${extension.name}"
+        }.simpleName
+        val messageName = extension.containingType.value.name
+        val typeNonNull = extension.typeFqNameNonNullable()
+        complexProperty(
+            name = descriptorName,
+            type = "%T<%T, ".scoped(
+                FqName.RpcClasses.InternalExtensionDescriptor,
+                messageName,
+            ).merge(typeNonNull) { first, typeNonNull -> "$first $typeNonNull>" },
```
Mr3zee, line 2257: and here

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -2012,12 +2201,226 @@ class ModelToProtobufKotlinCommonGenerator(
...
+    }
+
+    private fun CodeGenerator.generateExtensionDescriptor(field: FieldDeclaration) {
+        when (val type = field.type) {
+            is FieldType.IntegralType -> generateScalarExtensionDescriptor(field, type)
+            is FieldType.Enum -> generateEnumExtensionDescriptor(field, type)
+            is FieldType.Message -> generateMessageExtensionDescriptor(field, type)
+            is FieldType.List -> generateRepeatedExtensionDescriptor(field, type)
+            is FieldType.Map -> error("Extensions can't be of type Map")
+            is FieldType.OneOf -> error("Extensions can't be of type OneOf")
```
Mr3zee, line 2271: Never? Then you can remove todoos from tests above

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -2012,12 +2201,226 @@ class ModelToProtobufKotlinCommonGenerator(
...
+
+    private fun CodeGenerator.generateNonRepeatedExtensionDescriptor(
+        field: FieldDeclaration,
+        type: FieldType,
+    ) {
+        when (type) {
+            is FieldType.IntegralType -> generateScalarExtensionDescriptor(field, type)
+            is FieldType.Enum -> generateEnumExtensionDescriptor(field, type)
+            is FieldType.Message -> generateMessageExtensionDescriptor(field, type)
+            is FieldType.List -> error("Nested repeated extensions are not supported: ${field.name}")
```
Mr3zee, line 2365: By us or in protobuf? If by us, please, create a ticket

tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf_test_messages/edition2023/_rpc_internal/TestMessagesEdition2023.kt
```diff
@@ -6100,3 +6437,37 @@ fun TestAllTypesEdition2023.NestedEnum.Companion.fromNumber(number: Int): TestAl
...
+        InternalExtensionDescriptor.message(
+            fieldNumber = 121,
+            name = "groupliketype",
+            extendee = TestAllTypesEdition2023::class,
+            valueType = GroupLikeType::class,
+            default = { GroupLikeTypeInternal() },
+            asInternal = { it.asInternal() },
+            encodeWith = { value, encoder, config -> value.asInternal().encodeWith(encoder, config) },
+            decodeWith = { value, decoder, config -> GroupLikeTypeInternal.decodeWith(value.asInternal(), decoder, config) },
+        )
```
Mr3zee: cool generation!

tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf_test_messages/edition2023/TestMessagesEdition2023.ext.kt
```diff
@@ -287,3 +289,34 @@ interface ForeignMessageEdition2023Presence {
...
+val TestAllTypesEdition2023.delimitedExt: GroupLikeType get() = asInternal().getExtensionValue(TestMessagesEdition2023KtExtensions.delimitedExt) ?: TestMessagesEdition2023KtExtensions.delimitedExt.defaultValue.value
+
+val TestAllTypesEdition2023.Companion.delimitedExt: ProtoExtensionDescriptor<TestAllTypesEdition2023, GroupLikeType> get() = TestMessagesEdition2023KtExtensions.delimitedExt
+
+val TestAllTypesEdition2023Presence.hasDelimitedExt: Boolean get() = (this as InternalPresenceObject).hasExtension(TestMessagesEdition2023KtExtensions.delimitedExt)
+
+var TestAllTypesEdition2023.Builder.delimitedExt: GroupLikeType
+    get() = asInternal().getExtensionValue(TestMessagesEdition2023KtExtensions.delimitedExt) ?: TestMessagesEdition2023KtExtensions.delimitedExt.defaultValue.value
+    set(value) { asInternal().setExtensionValue(TestMessagesEdition2023KtExtensions.delimitedExt, value) }
+
```
Mr3zee, line 322: Should we move extensions to .ext files?

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalExtensionDescriptor.kt
Jozott00: You mean to enable reflection, so that all messages have a descriptor?

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: Non-user-facing messages are pseudo message classes, that are only generated for encoding/decoding purposes, but then transformed into an other representation. This is the case for map entries, which are a proto message type which is transformed into a normal Kotlin map, so a message never really holds such entries. I will add a comment.

Jozott00: Also, I will create an issue, because we should probably use `!isMapEntry` instead. When I created the `isUserFacing` extension, I thought there might be more cases then map entries, but it turned out to not be the case.

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: Ah yes thanks! It was previously more complex so this was a leftover

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalExtensionDescriptor.kt
Mr3zee: I mean, looks like the functions line `string()` define a generic way to work with a field, which is not tailored only to extension fields. Maybe we can reduce codegen size by reusing these functions for regualar fields too

Jozott00: Yeah, that’s essentially what we’d do if/when we implement descriptors. However, this might also affect performance since we’ll have more indirections. I’m not entirely satisfied with those extension descriptors yet and would put more effort into designing proper ones when we decide to implement descriptors for regular messages as well.

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: Yeah, protoc would fail

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: By protobuf, because this would be a `repeated repeated string field = 1`

tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf_test_messages/edition2023/TestMessagesEdition2023.ext.kt
Jozott00: They are in `.ext` files 👀

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: ahh thanks!

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalMessage.kt
Jozott00: Ups, yeah while writing the comment I noticed that I was wrong, fixed it and forgot about the comment ^^

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/ProtoExtension.kt
Jozott00: true, I meant "proto compiler plugin", so will fix it to `protoc plugin`

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/ProtoExtension.kt
Jozott00: Alright. We also have a `buildGrpcMetadata` function. Should we rename it as well?

Jozott00: And should I leave it as a function or expose an actual constructor on `ProtoExtensionRegistry` that takes the builder lambda?

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/ProtoExtension.kt
Jozott00: yeah, I will replace all of them in a follow up PR, I think we decided to use `Proto` instead of `Protobuf` in all cases.

protobuf/protobuf-api/src/commonTest/kotlin/kotlinx/rpc/protobuf/test/ProtobufSizeCalculationTest.kt
Jozott00: This was because the `stringExtension` is not real extension (not coming from a `.proto` file) so the test uses the internal API to build it ad-hoc to test the string proto type for extensions

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: True, for the generated code its on the same line. I will move the equals sign up in the comment

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: I was already pretty tired, apparently 👀

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: I think in this case actually not. `typeNonNull` is a `ScopedFormattedString`, and we can only pass FqName to `scoped` right?

protobuf/protobuf-api/src/commonTest/kotlin/kotlinx/rpc/protobuf/test/ProtoExtensionTest.kt
Jozott00: I am not sure if I understand what you mean 😅

protobuf/protobuf-wkt/src/commonMain/generated-code/kotlin-multiplatform/com/google/protobuf/kotlin/_rpc_internal/Any.kt
Jozott00: This is because of (from the generator source code)
```
            // this property is required to implement the InternalPresenceObject interface
            // for every possible case: if we have a nested message type with the same name
            // like MyClass1Internal.StringInternal.StringInternal, we cannot use "this@StringInternal"
            // as it is ambiguous.
            // to workaround this, we add the "_owner" property to the internal message, that can be
            // used by the presence object to reference its message.
```

protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/model/model.kt
Jozott00: I can rename it to `canHaveExtensions` or `hasExtensionRange`. What do you prefer ?

protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/CodeGenerator.kt
Jozott00: You mean the number of arguments or the total length of the call code?

protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/CodeGenerator.kt
Jozott00: True, the initial signature was in conflict with the original `property`, but now it works fine actually

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Mr3zee: Oh, you are right

tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf_test_messages/edition2023/TestMessagesEdition2023.ext.kt
Mr3zee: Apparently I'm also pretty tired😅

protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/CodeGenerator.kt
Mr3zee: Total length, yep

protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/model/model.kt
Mr3zee: has sounds good

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/ProtoExtension.kt
Mr3zee: Yeah, let's also rename. Builder functions are fine I think

protoc-gen/common/src/test/kotlin/kotlinx/rpc/protoc/gen/test/CodeGeneratorTest.kt
Jozott00: Such "complex" properties are often long and span over multiple lines, so I thought a line-break would be the best:
```
val repeatedInt32: InternalExtensionDescriptor<ExtensionBase,  List<Int>> =
    InternalExtensionDescriptor.packedRepeated(
        elementDescriptor =
            InternalExtensionDescriptor.int32(
                fieldNumber = 14,
                name = "repeatedInt32",
                extendee = ExtensionBase::class,
            ),
    )
```

vs

```
val repeatedInt32: InternalExtensionDescriptor<ExtensionBase,  List<Int>> = InternalExtensionDescriptor.packedRepeated(
            elementDescriptor =
                InternalExtensionDescriptor.int32(
                    fieldNumber = 14,
                    name = "repeatedInt32",
                    extendee = ExtensionBase::class,
                ),
        )
```

protobuf/protobuf-api/src/commonTest/kotlin/kotlinx/rpc/protobuf/test/ProtoExtensionTest.kt
Mr3zee: For me it looks like line 47 and 50 do the same thing

protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/CodeGenerator.kt
Jozott00: For that I would have to resolve the `ScopeFormmatedString` values, no?

tests/protobuf-conformance/src/main/generated-code/kotlin-multiplatform/com/google/protobuf_test_messages/edition2023/TestMessagesEdition2023.ext.kt
Jozott00: 😬

protobuf/protobuf-api/src/commonTest/kotlin/kotlinx/rpc/protobuf/test/ProtoExtensionTest.kt
Jozott00: Oh, in 47 for, we just register the descriptor in the registry and in `50` we get it out of the registry and check if it was there.

Jozott00: I guess the name `allExtensionsForMessage` isn't very clear. Do you have a better name in mind? It gets all extensions that were registered for a specific message.

Mr3zee: Ooooh
Then just add a get in from of the name and should be fine

protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/CodeGenerator.kt
Mr3zee: 🤔 
Maybe, I don't remember exactly. I did it somehow in class generation. Maybe it is wrong actually, in that case we can forget about it for a while

protoc-gen/common/src/test/kotlin/kotlinx/rpc/protoc/gen/test/CodeGeneratorTest.kt
Mr3zee: This can be a parameter

Jozott00: True 😅

protobuf/protobuf-api/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalExtensionDescriptor.kt
```diff
@@ -12,41 +12,47 @@ import kotlin.reflect.cast
 import kotlin.reflect.safeCast
 
 @InternalRpcApi
+/**
+ * Internal runtime descriptor for a protobuf extension field.
+ */
```
Mr3zee: Please swap annotation and the comment

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -249,7 +249,7 @@ class ModelToProtobufKotlinCommonGenerator(
                     }
                 }
 
-                property(
+                this.property(
```
Mr3zee: These `this.` calls looks like something IDEA did, are they needed?

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -863,9 +854,9 @@ class ModelToProtobufKotlinCommonGenerator(
 
     private fun CodeGenerator.generatePublicPresenceGetter(declaration: MessageDeclaration) {
         if (!declaration.isUserFacing) return
-        if (!declaration.hasPresenceFields && !declaration.mayHaveExtensions) return
+        if (!declaration.hasPresenceFields && !declaration.hasExtensionRange) return
 
-        property(
+        this@generatePublicPresenceGetter.property(
```
Mr3zee: this too

protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
Jozott00: True, I will remove them

PR #589, opened 2026-03-02
pb: Add MARSHALLER and DESCRIPTOR object to group messages

Jozott00:
**Subsystem**
Protobuf

**Problem Description**
Group messages were not annotated with `@GeneratedProtoMessage` because they didn't have a marshaller and descriptor object.

**Solution**
Now each group message also has a `MARSHALLER` object that allows encoding and decoding of a standalone group message. The encoding and decoding works like for normal functions, so no `GROUP_START` and `GROUP_END` tag are written or read. 
While this isn't something that the protobuf spec intends, it doesn't violate any specs and allows us to annotate it with `@GeneratedProtoMessage`.

### DIFF
--- a/protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
+++ b/protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
@@ -770,8 +770,6 @@ class ModelToProtobufKotlinCommonGenerator(
 
     private fun CodeGenerator.generateMarshallerObject(declaration: MessageDeclaration) {
         if (!declaration.isUserFacing) return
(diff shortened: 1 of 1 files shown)


protoc-gen/protobuf/src/main/kotlin/kotlinx/rpc/protoc/gen/ModelToProtobufKotlinCommonGenerator.kt
```diff
@@ -774,7 +771,7 @@ class ModelToProtobufKotlinCommonGenerator(
     private fun CodeGenerator.generateMarshallerObject(declaration: MessageDeclaration) {
         if (!declaration.isUserFacing) return
         // the MARSHALLER object is not necessary for groups, as they are inlined messages
-        if (declaration.isGroup) return
+        // if (declaration.isGroup) return
```
Mr3zee: Can be safely deleted

PR #590, opened 2026-03-02
Unittest proto: First iteration

Mr3zee:
**Subsystem**
Protoc-gen

**Problem Description**
There are Google's protobuf tests called unittest. We want to use them to provision our generator.

**Solution**
Add configs to generate against these files

**Warning**
This is not complete test set. Some tests are just failing because our generator is not capable enough yet. I'll fix them in later PRs

### DIFF
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -67,6 +67,22 @@ In the root of the project you should create a `local.properties` file with the 
 sdk.dir=/Users/<user>/Library/Android/sdk
 ```
 
+## YouTrack MCP
+
+The project includes a `.mcp.json` that connects Claude Code to YouTrack via MCP.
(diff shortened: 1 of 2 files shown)


docs/environment.md
```diff
@@ -67,6 +67,22 @@ In the root of the project you should create a `local.properties` file with the
 sdk.dir=/Users/<user>/Library/Android/sdk
 ```
 
+## YouTrack MCP
```
Jozott00, line 70: Nice, what can I do with it?

Mr3zee: Basically, work with YT from agents

Mr3zee: You can build the whole workflow: find issues -> create reproducer -> try fix -> create PR

PR #635, opened 2026-03-25
pb+grpc: Prebuild c-interop binaries

Jozott00:
**Subsystem**
Protobuf/Native, gRPC/Native, Gradle Setup

**Issues**
Closes KRPC-528

**Problem**
Currently, the protobuf and grpc cinterop klibs are built as part of the protobuf-api and grpc-core modules. This means that during configuration in the IDE, all of them have to be build (for each target). 
Additionally, tasks like dokka that have to be configured for all targets fail on Linux, as Apple targets can't be configured on that host platform.
Furthermore, the prebuilt grpc archives are currently included in the repository, which makes it hard to version them and causing the git repository to become large in size.

**Solution**
This PR solves these problems by restructuring the project setup and publication for grpc/protobuf cinterop + shim parts.
1. The grpc C core artifacts are no longer compiled into a single large static archive. Instead, all static archives (`.a`) are collected and attached to the klib. This change is beneficial because it minimizes the risk of invalid symbol duplication during the creation of a large static archive. Moreover, the previously used bazel rule `cc_static_library` is an experimental feature.
2. The grpc C core artifacts are no longer committed to the git project but instead published to the `grpc` space repository as `zip` maven artifacts. These artifacts match the exact version of the built grpc C core version, enabling versioning and reducing the size of the git repository. The `grpc-c-prebuilt` project resides under `native-deps/grpc-c-prebuilt` and is an independent gradle project, not linked to `kotlinx-rpc`. 
3. The `grpc-core` module is no longer integrating the cinterop task, but instead depends on an separately published klib that integrates the cinterop declarations, shim and static archives of grpc. This project lives under `native-deps/shims/grpc` which is a submodule of the `native-deps/shims` standalone project. It is important to know that
    1. It is published to maven-central (once gRPC support is stable) as it is a real transitive dependency of `grpc-core`
    2. It's implementation and exposed API is unstable, with no backward compatibility guarantees, and therefore should never be consumed as library itself.
    3. To make the above clear to the user, the cinterop generated kotlin declarations in the klib are patched before publication, to be annotated with `@InternalNativeRpcApi`. 
    4. It includes all static libraries that were published by `native-deps/grpc-c-prebuilt`.
    5. It's versioning is `<grpc-upstream-version>-<incremental-shim-version>`. 
    6. It is published entirely independently of `kotlinx-rpc`.
4. As the `protobuf-api` module has the same issues as the `grpc-core` moduel, we apply the same approach  as for the `shims/grpc` project but for `protobuf`. So there is also a `native-deps/shims/protobuf` module that behaves the same as the `shims/grpc` module. Note that we might remove this module and dependency, if we implement our own pure Kotlin wire encoding.

### DIFF
--- a/grpc/grpc-core/build.gradle.kts
+++ b/grpc/grpc-core/build.gradle.kts
@@ -86,7 +86,7 @@ kotlin {
 
         nativeMain {
             dependencies {
-                // required for status.proto
+                // TODO: Remove this dependency once we remove the protobuf-shim dependency (KRPC-540)
                 implementation(projects.protobuf.protobufApi)
                 implementation(libs.kotlinx.rpc.grpc.core.shim)
             }


gradle-conventions-settings/src/main/kotlin/conventions-repositories.settings.gradle.kts
```diff
@@ -207,6 +207,9 @@ gradle.rootProject {
 
         buildscript {
             repositories {
+                // Temporary local development fallback while grpc-shim is still consumed from mavenLocal.
+                mavenLocal()
```
Mr3zee: Can we publish it to buildRepo? (the one in `<root>/build/repo`)

gradle-conventions/src/main/kotlin/util/bazel.kt
```diff
@@ -0,0 +1,160 @@
...
+
+fun Project.konanHomeProvider(): Provider<String> = providers.provider { findKonanHome() }
+
+fun Project.requireGradleProperty(name: String): String = providers.gradleProperty(name).orNull
+    ?: error("Missing $name in ${layout.projectDirectory.file("gradle.properties").asFile.absolutePath}")
+
+fun Project.registerCheckBazelTask(name: String = "checkBazel"): TaskProvider<Exec> =
+    tasks.register<Exec>(name) {
+        group = "verification"
+        description = "Checks that Bazel or Bazelisk is available on PATH"
```
Mr3zee: should probably be just Bazelist otherwise we might face unstable build with different Bazel versions

gradle-conventions/src/main/kotlin/util/publication.kt
```diff
@@ -92,19 +92,29 @@ private fun File.fixMetadata(projectName: String, projectVersion: String) {
 // from: "name": "krpc-client-metadata-0.2.0.jar",
 // to: "name": "kotlinx-rpc-krpc-client-metadata-0.2.0.jar",
 private fun String.updateFilesNameField(projectName: String): String {
+    val prefixedName = if (projectName.startsWith(KOTLINX_RPC_PREFIX)) {
+        projectName
```
Mr3zee: In this case the update is not needed at all

gradle-conventions/src/main/kotlin/conventions-publishing.gradle.kts
```diff
@@ -11,6 +11,10 @@ import util.tasks.ValidatePublishedArtifactsTask
 
 val isGradlePlugin = project.name == "gradle-plugin"
 val isProtocGen = project.rootProject.name.startsWith(PROTOC_GEN)
+val shouldPublishEmptyJavadoc = project.findProperty("kotlinx.rpc.publishing.emptyJavadoc")
+    ?.toString()
+    ?.toBooleanStrictOrNull()
+    ?: true
```
Mr3zee: Why it might be false?

grpc/grpc-core/build.gradle.kts
```diff
@@ -17,6 +14,12 @@ plugins {
     alias(libs.plugins.serialization) // for tests
 }
 
+repositories {
+    mavenLocal()
+}
+
```
Mr3zee: If you have in gradle-settings-conventions then you don't need it here

grpc/grpc-core/build.gradle.kts
```diff
@@ -17,6 +14,12 @@ plugins {
...
+val publishedGrpcCoreShimVersion = "1.74.1-1"
```
Mr3zee: Can we put that into the libs.versions.toml file and make sure it is synced with the grpc version by a GH workflow (we can run it only on the versions file changes)

grpc/grpc-core/build.gradle.kts
```diff
@@ -91,6 +94,7 @@ kotlin {
             dependencies {
                 // required for status.proto
                 implementation(projects.protobuf.protobufApi)
+                implementation("org.jetbrains.kotlinx:kotlinx-rpc-grpc-core-shim:$publishedGrpcCoreShimVersion")
```
Mr3zee: Also can go into the versions file

native-deps/bazel-support/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Used by:
+- `native-deps/grpc`
+- `native-deps/grpc-shim`
+- `native-deps/protobuf-shim`
+
+Kotlin/Native toolchain notes:
+- Linux target builds use the Kotlin/Native toolchain via the shared Bazel toolchain in `toolchain/`.
+- `toolchain/konan_llvm_bundles.json` pins LLVM bundle names per Kotlin compiler version.
+- Do not manually patch toolchain internals when Kotlin changes; refresh the mapping instead.
```
Mr3zee, line 17: Do we have a workflow to check that?

native-deps/bazel-support/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+Updating Kotlin compiler support:
+- Run `./toolchain/precompute_konan_llvm_bundles.py` from this directory to refresh `konan_llvm_bundles.json`.
+- The script reads upstream Kotlin Native metadata and updates the host-to-LLVM bundle mapping used by native-deps builds.
+- CI verifies that the mapping matches the current project Kotlin compiler version.
+
+Troubleshooting:
```
Mr3zee, line 24: Great to have that section!

native-deps/grpc-shim/core/build.gradle.kts
```diff
@@ -0,0 +1,341 @@
...
+val grpcCoreInteropDefTemplate = layout.projectDirectory.file("src/nativeInterop/cinterop/grpcCoreInterop.def")
+val grpcShimTargets = nativeDependencyTargets
+val grpcShimModuleFile = layout.projectDirectory.file("MODULE.bazel").asFile
+
+// cinterop does not load our compiler plugin on this build path, so grpc-shim uses a dedicated internal
+// KLIB patcher to add the opt-in marker after cinterop and before publication.
+val internalNativeRpcApiPatcherProject = project(":klib-patcher")
+val internalNativeRpcApiPatcherJar = internalNativeRpcApiPatcherProject.layout.buildDirectory.file(
+    "libs/${internalNativeRpcApiPatcherProject.name}-${project.version}.jar",
+)
```
Mr3zee: This script file looks like a great candidate to be a plugin in the conventions directory to keep the build files simpler

native-deps/grpc-c-prebuilt/build_archives.sh
```diff
@@ -0,0 +1,164 @@
...
+
+ARCHIVE_TUPLES="$archive_tuples" ARCHIVE_DST="$DST" ARCHIVE_LLVM_AR="$LLVM_AR" python3 - <<'PY'
+import ast
+import os
+import shutil
+import subprocess
+from pathlib import Path
+from typing import List
+
+tuples = ast.literal_eval(os.environ["ARCHIVE_TUPLES"])
```
Mr3zee: This seems like it can be python file 😄

native-deps/grpc/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+Shared toolchain notes:
+- Manual Bazel/Konan troubleshooting and LLVM bundle maintenance live in [../bazel-support/README.md](/Users/jozott/development/jetbrains/kotlinx-rpc/native-deps/bazel-support/README.md).
+
+Useful tasks:
+- `../../gradlew -p native-deps/grpc buildGrpcHeaders`
+- `../../gradlew -p native-deps/grpc packageGrpcHeaders`
+- `../../gradlew -p native-deps/grpc buildGrpcIosArm64`
+- `../../gradlew -p native-deps/grpc buildAllGrpcBundles`
+- `../../gradlew -p native-deps/grpc publishAllPublicationsToBuildRepoRepository`
```
Mr3zee: If it is ../../gradlew, why there is gradlew in the local root too?

native-deps/shims/klib-patcher/src/main/java/kotlinx/rpc/nativedeps/tooling/KlibPatcher.java
```diff
@@ -0,0 +1,387 @@
...
+import java.util.zip.ZipEntry;
+import java.util.zip.ZipOutputStream;
+
+/**
+ * Patches produced cinterop KLIB metadata so declarations in one target package require explicit opt-in.
+ *
+ * <p>This is shared internal native-deps build tooling used by shim builds such as grpc-shim and
+ * protobuf-shim. It is intentionally parameter-driven so the fragile KLIB metadata rewrite logic lives in one place.
+ */
+public final class KlibPatcher {
```
Mr3zee: Any particular reason why it is in Java?
Also, this seems like a very comlicated way to add an annotation to all declarations

native-deps/protobuf-shim/annotation/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalNativeRpcApi.kt
```diff
@@ -0,0 +1,27 @@
...
+)
+@Retention(AnnotationRetention.BINARY)
+@Target(
+    AnnotationTarget.CLASS,
+    AnnotationTarget.CONSTRUCTOR,
+    AnnotationTarget.FUNCTION,
+    AnnotationTarget.PROPERTY,
+    AnnotationTarget.TYPEALIAS,
+)
+public annotation class InternalNativeRpcApi
```
Mr3zee: This one is probably `InternalNativeProtobufApi`

native-deps/protobuf-shim/core/build.gradle.kts
```diff
@@ -0,0 +1,215 @@
...
+ * 2. Running cinterop against the published protowire headers and static library.
+ * 3. Patching the produced cinterop KLIB so the generated declarations require
+ *    explicit opt-in via `InternalNativeRpcApi`.
+ */
+plugins {
+    alias(libs.plugins.kotlin.multiplatform)
+    id("conventions-publishing")
+}
+
+val protobufVersion = providers.gradleProperty("protobufVersion").get()
```
Mr3zee: A lot from this build file looks the same as in the grpc one, can they share the logic and store it in conventions?

native-deps/shims/tests/build.gradle.kts
```diff
@@ -0,0 +1,88 @@
...
+
+data class HostTarget(
+    val declaration: String,
+    val compileTask: String,
+    val publicationTaskSuffix: String,
+    val publicationSuffix: String,
+)
+
+fun currentHostTarget(): HostTarget {
+    val os = System.getProperty("os.name").lowercase()
```
Mr3zee, line 44: This file too looks like the one from grpc

native-deps/protobuf-shim/README.md
```diff
@@ -0,0 +1,21 @@
...
+- `../klib-patcher/` contains the shared internal metadata patcher used only while building `core`
+- `tests/` contains TestKit-based fixture tests that publish protobuf-shim artifacts locally and compile throwaway consumer projects
+
+Shared toolchain notes:
+- Manual Bazel/Konan troubleshooting and LLVM bundle maintenance live in [../bazel-support/README.md](/Users/jozott/development/jetbrains/kotlinx-rpc/native-deps/bazel-support/README.md).
+
+Useful tasks:
+- `../../gradlew -p native-deps/protobuf-shim :core:publishAllPublicationsToVerificationRepository`
+- `../../gradlew -p native-deps/protobuf-shim :annotation:publishAllPublicationsToVerificationRepository`
+- `../../gradlew -p native-deps/protobuf-shim :tests:test`
```
Mr3zee: Same question about ../../gradlew

It seems like the whole native-deps can be one gradle project with submodules/included builds.

protobuf/protobuf-api/build.gradle.kts
```diff
@@ -7,15 +7,19 @@
...
     alias(libs.plugins.conventions.kmp)
     alias(libs.plugins.kotlinx.rpc)
 }
 
+repositories {
+    mavenLocal()
+}
+
 val globalRootDir: String by project.extra
+val publishedProtobufShimVersion = "31.1-1"
```
Mr3zee: Same thing about versions file

gradle-conventions/src/main/kotlin/conventions-publishing.gradle.kts
Jozott00: The shims never contain any java docs, so I thought it would be unnecessary to publish the javadocs artifact for them. I can still include them if you want.

gradle-conventions-settings/src/main/kotlin/conventions-repositories.settings.gradle.kts
Jozott00: This was a left over of testing during development. It is removed now.

gradle-conventions/src/main/kotlin/conventions-publishing.gradle.kts
Mr3zee: This is a Maven Central hard requirement, so we need to publish them

Jozott00: Ahh ok good to know, will remove it then

native-deps/bazel-support/README.md
Jozott00: yes, there is verify-konan-llvm-bundles.yml

native-deps/grpc-shim/core/build.gradle.kts
Jozott00: This is extracted now

native-deps/shims/klib-patcher/src/main/java/kotlinx/rpc/nativedeps/tooling/KlibPatcher.java
Jozott00: I switched it back to Kotlin and compilation did not went OOM anymore. Not sure why it was the case before.

native-deps/grpc-c-prebuilt/build_archives.sh
Jozott00: Yeah, at this point...  I will chnage it to python

native-deps/grpc-c-prebuilt/gradle.properties
```diff
@@ -0,0 +1,6 @@
+#
+# Copyright 2023-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+#
+
+grpcVersion=1.74.1
```
Mr3zee: Can it be synced with libs.versions.toml?

native-deps/shims/annotation/src/commonMain/kotlin/kotlinx/rpc/protobuf/internal/InternalNativeProtobufApi.kt
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright 2023-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.rpc.protobuf.internal
```
Mr3zee: Let's make it `kotlinx.rpc.protobuf.internal.shim`, same for grpc. Packages must be unique across all modules

Mr3zee: So any standalone published artifact should have a unique package

native-deps/shims/grpc/src/nativeMain/kotlin/KgrpcNative.kt
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright 2023-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.rpc.grpc.nativedeps
+
+internal object KgrpcNative
```
Mr3zee, line 13: Please add 'don't remove' comment with a note that this keeps the build working

native-deps/shims/protobuf/src/nativeMain/kotlin/ProtobufShimNative.kt
```diff
@@ -0,0 +1,7 @@
+/*
+ * Copyright 2023-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.rpc.protobuf.nativedeps
+
+internal object ProtobufShimNative
```
Mr3zee, line 13: same here

native-deps/shims/tests/src/test/kotlin/kotlinx/rpc/protobuf/internal/ProtobufShimFixtureTest.kt
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright 2023-2026 JetBrains s.r.o and contributors. Use of this source code is governed by the Apache 2.0 license.
+ */
+
+package kotlinx.rpc.protobuf.internal
+
+import kotlinx.rpc.nativedeps.shims.NativeShimFixtureTestSupport
+
+class ProtobufShimFixtureTest : NativeShimFixtureTestSupport() {
```
Mr3zee, line 9: Do we have configs to run this on CI?

native-deps/shims/build.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+                    url = uri(rootProject.layout.buildDirectory.dir("verification-repo"))
+                }
+            }
+        }
+    }
+}
+
+project(":kotlinx-rpc-grpc-core-shim").version = "$grpcVersion-$grpcShimVersion"
+project(":kotlinx-rpc-protobuf-shim").version = "$protobufVersion-$protobufShimVersion"
+project(":kotlinx-rpc-native-shims-annotation").version = annotationVersion
```
Mr3zee: Potentially breaks project isolation in the future, can it be moved into dedicated projects?

native-deps/shims/gradle.properties
```diff
@@ -0,0 +1,19 @@
...
+# Version numbers for the gRPC shim (<grpcVersion>-<grpcShimVersion>).
+grpcVersion=1.74.1
+grpcShimVersion=2
+
+# Version numbers for the Protobuf shim (<protobufVersion>-<protobufShimVersion>).
+protobufVersion=31.1
+protobufShimVersion=2
+
+# Version numbers for the annotation shim.
+annotationVersion=0.1.0
```
Mr3zee: Same question about versions

native-deps/grpc-c-prebuilt/gradle.properties
Jozott00: Not sure if we should, as it's essentially completely independent, because before we can use it in the kotlinx-rpc project, we must build and publish the `grpc-c-prebuilt` archives and then update and publish the `shims/grpc`.

native-deps/shims/tests/src/test/kotlin/kotlinx/rpc/protobuf/internal/ProtobufShimFixtureTest.kt
Jozott00: Yes, under `builds/native-deps/`. Those tests are also release build dependencies.

native-deps/shims/build.gradle.kts
Jozott00: So you mean that the grpc-shim version is set in the `grpc/build.gradle.kts`? Yeah, I will change that.

PR #644, opened 2026-04-03
protoc-gen: Refactor optional field generation (KRPC-262)

Jozott00:
**Subsystem**
Protobuf/protoc-gen

**Problem Description**
Currently, optional fields that lack a user-defined default value are generated as nullable properties, with the exception of sub-message fields. 

This implementation differs from the standard protobuf approach, which handles all optional fields as non-nullable. This allows users to quickly access default values for non-set subfields. Furthermore, in the edition 2023+ version, where all fields are typically optional by default, managing nullable properties becomes cumbersome.  

**Solution**
This PR aligns the implementation with official Protobuf implementations, ensuring that all fields (except one-of fields that lack a default value) are non-nullable. However, to maintain Kotlin’s idiomatic nullable handling, users can enable additional nullable getter generation, which produces
```kt
// in MyProto.ext.kt
val MyMessage.myFieldOrNull: FieldType?
```

Additionally, `MyMessage.Builder.clear<Field>()` functions are generated for all fields that are presence tracked. This enables users to clear the original set value in the copy block 
```kt
val copy = message.copy {
  clearMyField()
}
```


compiler-plugin/compiler-plugin-k2/src/main/kotlin/kotlinx/rpc/codegen/FirGenerationKeys.kt
```diff
@@ -49,3 +49,9 @@ internal object FirGeneratedProtoMessageBuilderPropertyKey : GeneratedDeclaratio
         return "FirGeneratedProtoMessageBuilderPropertyKey"
     }
 }
+
+internal object FirGeneratedProtoMessageBuilderFunctionKey : GeneratedDeclarationKey() {
+    override fun toString(): String {
+        return "FirGeneratedProtoMessageBuilderFunctionKey"
+    }
+}
```
Jozott00, line 57: Not sure if this is actually needed, or if we could just use the PropertyKey

grpc/grpc-core/src/desktopTest/kotlin/kotlinx/rpc/grpc/test/integration/TestReferenceService.kt
eminasljivic: These todos can now be resolved, right?

PR #662, opened 2026-04-12
Work around native segfault in callables property dispatch

andrewparmet:
**Subsystem**
gRPC server, gRPC core

**Problem Description**
`descriptor.callables` segfaults on Kotlin/Native when consuming published kotlinx-rpc artifacts, making `GrpcServerImpl.registerService()` crash for any `@Grpc` service. See https://github.com/Kotlin/kotlinx-rpc/issues/663.

`descriptor.getCallable(name)`, which reads from the same backing field, works fine.

**Solution**
Avoid `descriptor.callables` in `GrpcServerImpl.getDefinition()` by iterating `delegate.methodNames` and using `descriptor.getCallable()` instead.

Adds `GrpcServiceDelegate.methodNames` to expose the method name set (the underlying `methodDescriptorMap` is private).


Mr3zee:
Thanks for the workaround PR! Fortunately, the proper fix was landed in https://github.com/Kotlin/kotlinx-rpc/pull/668, and so I'll close this one

PR #696, opened 2026-04-22
grpc: Release grpc_call deterministically in NativeClientCall (KRPC-586)

Mr3zee:
### Subsystem

grpc-client (native)

### Problem

YouTrack: [KRPC-586](https://youtrack.jetbrains.com/issue/KRPC-586)

### Solution

`NativeClientCall` was relying on a Kotlin/Native GC cleaner as the only mechanism to call `grpc_call_unref` on the application's +1 ref. Kotlin/Native GC timing is not coupled to the `rt.close()` chain that ends in `grpc_shutdown()`, so on linuxX64 the library-shutdown could execute while `grpc_call` objects were still owned by the application — a direct violation of grpc-core's documented precondition in `grpc/grpc.h`. Once grpc-core began internal teardown, the pending cleaner's later `grpc_call_unref` touched freed state → SIGSEGV.

The fix introduces deterministic release of `grpc_call` at the point where the call becomes logically dead (`tryToCloseCall`, right after `listener.onClose` fires), using a `ResourceGuard` (the same pattern already used for `grpc_channel_destroy` in `NativeManagedChannel`, now promoted to `internal` so both files share it). The GC cleaner remains as a fallback.

Three additional protections keep the deterministic unref race-free with concurrent user operations:

- `runBatch` re-checks `closed` after `beginOp()`. The first check is the fast path; once `inFlight > 0`, `tryToCloseCall` cannot fire — so if the second check reads `closed=false`, `raw` is safe to use until the matching `endOp`. The SC-ordering contract between the `inFlight.incrementAndGet` and the `closed.value` read is what makes this work; a comment in `tryToCloseCall` calls this out explicitly.
- `cancelInternal` now runs under its own `beginOp`/`endOp` pair with a `closed.value` short-circuit. Without this, a user-side `cancel()` could race with `tryToCloseCall`'s unref and UAF on `grpc_call_cancel_with_status`.
- The `CQShutdown` / `SubmitError` branches of `runBatch` call `cancelInternal` **before** the outer `endOp`, rather than after. This keeps `inFlight` non-zero across `cancelInternal`, so there is no transient window in which another thread could see `inFlight == 0` and unref `raw` from under us.

Server-side (`NativeServerCall`) is deliberately not changed in this PR:

1. The failing CI tests do not reach the server (TLS handshake / compression check / interceptor cancel all fail on the client).
2. Server user APIs (`sendMessage`, `close`) lack the `beginOp`/`endOp` fence that makes the deterministic client-side unref safe against concurrent user ops — adding an eager unref at `finalize` without that fence would introduce a new UAF risk. Server-side deterministic release will need a separate design.

Follow-ups filed: [KRPC-588](https://youtrack.jetbrains.com/issue/KRPC-588) (`rawCallCredentials` still has the same GC-only pattern; scoped-out because it was not part of the crash signature), [KRPC-589](https://youtrack.jetbrains.com/issue/KRPC-589) (pre-existing crash in `tryToCloseCall` if the call is cancelled before `start()` sets the listener — unchanged by this PR, but surfaced during review).

Verified locally: `:grpc:grpc-core:jvmTest` (159/159) + `:grpc:grpc-core:macosArm64Test` (159/159), `:grpc:grpc-client:detekt`, `:grpc:grpc-client:checkLegacyAbi` all pass. linuxX64 requires CI to verify.

---

> [!NOTE]
> Fully autonomous AI-generated PR — no human reviewed the code before submission.
> Problem analysis and root cause details: [KRPC-586](https://youtrack.jetbrains.com/issue/KRPC-586)

### DIFF
--- a/grpc/grpc-client/src/nativeMain/kotlin/kotlinx/rpc/grpc/client/internal/NativeClientCall.kt
+++ b/grpc/grpc-client/src/nativeMain/kotlin/kotlinx/rpc/grpc/client/internal/NativeClientCall.kt
@@ -77,9 +77,16 @@ internal class NativeClientCall<Request, Response>(
     private val coroutineContext: CoroutineContext,
 ) : ClientCall<Request, Response>() {
 
+    // grpc_shutdown() requires all application-owned grpc objects to be destroyed before it runs
+    // (grpc/grpc.h). Release the application's +1 grpc_call ref deterministically in
+    // tryToCloseCall; the cleaner is the fallback for cases where onClose never fires. KRPC-586.
+    private val rawGuard = ResourceGuard()
+
     @Suppress("unused")
-    private val rawCleaner = createCleaner(raw) {
-        grpc_call_unref(it)
+    private val rawCleaner = createCleaner(Pair(raw, rawGuard)) { (ptr, guard) ->
+        if (guard.released.compareAndSet(expect = false, update = true)) {
+            grpc_call_unref(ptr)
+        }
     }
 
     private val rawCallCredentials = callOptions.callCredentials.let {
@@ -102,6 +109,13 @@ internal class NativeClientCall<Request, Response>(
                 is Throwable -> {
                     cancelInternal(grpc_status_code.GRPC_STATUS_INTERNAL, "Call failed: ${it.message}")
                 }
+            }
+            // Fallback deterministic release for calls that never reached tryToCloseCall — e.g., a
+            // client interceptor threw before start() submitted any batch, so no
+            // RECV_STATUS_ON_CLIENT ever completes and closeInfo stays null. Without this, the
+            // grpc_call is owned past grpc_shutdown(). rawGuard blocks double-unref. KRPC-586.
+            if (rawGuard.released.compareAndSet(expect = false, update = true)) {
+                grpc_call_unref(raw)
             }
         }
     }
@@ -157,12 +171,26 @@ internal class NativeClientCall<Request, Response>(
      */
     private fun tryToCloseCall() {
         val info = closeInfo.value ?: return
+        // The `inFlight.value == 0` read and the `closed` CAS below are intentionally non-atomic
+        // together. A concurrent `beginOp` can increment inFlight after we read it here; runBatch's
+        // post-beginOp re-read of `closed` is the barrier that closes that window. Under SC atomics
+        // (atomicfu on K/N), if our CAS on `closed` is sequenced after the thread's
+        // `inFlight.incrementAndGet`, that thread's following `closed.value` load will observe
+        // `closed=true` and bail before touching raw. Do not weaken either ordering.
         if (inFlight.value == 0 && closed.compareAndSet(expect = false, update = true)) {
-            val lst = checkNotNull(listener) { internalError("Not yet started") }
             // allows the managed channel to join for the call to finish.
             callJob.complete()
-            safeUserCode("Failed to call onClose.") {
-                lst.onClose(info.first, info.second)
+            // Listener may be null if the call failed before start() (e.g., an interceptor throws
+            // mid-chain and the call is later cancelled via shutdownNow → markClosePending). No
+            // user observer to notify in that case; resources still need to be released below.
+            listener?.let { lst ->
+                safeUserCode("Failed to call onClose.") {
+                    lst.onClose(info.first, info.second)
+                }
+            }
+            // Deterministic grpc_call_unref — see rawGuard.
+            if (rawGuard.released.compareAndSet(expect = false, update = true)) {
+                grpc_call_unref(raw)
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Mr3zee:
## Internal code review

Two review agents reviewed the change. All errors and important issues flagged were fixed before pushing. Remaining notes below are pre-existing issues surfaced during review, filed as follow-ups and **not** in scope for this PR.

**Agent #1 (concurrency/lifecycle)**
- Important / ✅ fixed in this PR: non-obvious SC-ordering contract between `inFlight.incrementAndGet` and subsequent `closed.value` reads — added a comment to `tryToCloseCall` documenting why the `&&` in `inFlight.value == 0 && closed.CAS` is safe under the two-check pattern in `runBatch`.
- Important / ✅ fixed in this PR: `CQShutdown` / `SubmitError` branches of `runBatch` had a transient `inFlight == 0` window between `endOp()` and `cancelInternal()`. Swapped the order so the outer `beginOp` stays open across `cancelInternal`.
- Important / ⏭️ out of scope: `rawCallCredentials` is still released only via the GC cleaner. Same class of issue as KRPC-586 but for credentials rather than calls; narrower race window. Filed as [KRPC-588](https://youtrack.jetbrains.com/issue/KRPC-588).
- Suggestion / 🤷 not adopted: replace `rawGuard` with `closed` since both are always set together. Rejected to match the `ResourceGuard` pattern already in use for the channel.
- Suggestion / 🤷 pre-existing, not adopted: `cancel()` uses `closeInfo.value = ...` (direct write) rather than CAS. Intentional "user wins" semantics, unchanged by this PR.

**Agent #2 (design/pattern)**
- Important / ✅ fixed in this PR: use `ResourceGuard` instead of inlining `atomic(false)` + `Pair` capture. Promoted `ResourceGuard` from `private class` in `NativeManagedChannel.kt` to `internal class` so both files share it.
- Suggestion / ✅ fixed in this PR: comment redundancy between the field-level `rawGuard` doc and the in-method call site. Trimmed the in-method comment to a cross-reference.
- Suggestion / ⏭️ out of scope: `tryToCloseCall` crashes on `checkNotNull(listener)` if `callJob` is cancelled before `start()`. Pre-existing bug, this PR marginally widens the reachable window. Filed as [KRPC-589](https://youtrack.jetbrains.com/issue/KRPC-589).

Mr3zee:
### CI iteration 1 → 2 (scope expansion)

First CI run surfaced a same-class crash on a path the initial fix did not cover:

- TC Linux (53146): ✅ SUCCESS (1367 passed, 18 ignored)
- TC Apple (53147): ❌ FAILURE on `ClientInterceptorTest.throw during intercept - should fail with thrown exception[macosArm64]` with the same "Test running process exited unexpectedly" symptom

Confirms the user's hypothesis that the bug class is platform-independent (linuxX64 just loses the race more reliably). The specific path the initial fix missed:

When an interceptor throws **before** `start()` is called, the `NativeClientCall` is created (so the application owns a `grpc_call*`) but no `RECV_STATUS_ON_CLIENT` batch is ever submitted. `closeInfo` stays null, so `tryToCloseCall` never fires — and the initial fix's deterministic unref lived inside `tryToCloseCall`.

Extended the fix (force-pushed, now at `cd48fcd0`):

1. Added fallback `grpc_call_unref` in `callJob.invokeOnCompletion` (triggered by `shutdownNow()` cancelling the child job). `rawGuard` prevents any double-unref with the in-`tryToCloseCall` path for normal completions.
2. Made `tryToCloseCall` tolerate a null listener — the pre-existing `checkNotNull(listener)` crash (filed as KRPC-589) becomes newly reachable via the fallback-unref path when the call is cancelled after an interceptor failure. Rolled KRPC-589 into this PR as duplicate / fixed-by-this-PR.

Re-triggered: TC [53149](https://krpc.teamcity.com/buildConfiguration/Build_kRPC_Java_JS_WASM_and_Linux/53149) (Linux) + [53150](https://krpc.teamcity.com/buildConfiguration/Build_kRPC_Apple/53150) (Apple). GH Actions re-running on the new push.

Mr3zee:
## CI Report

### Passed

| Pipeline | Details |
|----------|---------|
| TC: Java, JS, WASM and Linux (Code) | [build #53149](https://krpc.teamcity.com/buildConfiguration/Build_kRPC_Java_JS_WASM_and_Linux/53149) — 1159 passed, 16 ignored (12m 6s) |
| TC: Apple (Code) | [build #53150](https://krpc.teamcity.com/buildConfiguration/Build_kRPC_Apple/53150) — 592 passed, 4 ignored (7m 45s) |
| GH: Check AI PR Safety | [run #24774040271](https://github.com/Kotlin/kotlinx-rpc/actions/runs/24774040271) |
| GH: Run Verification (×7) | [runs on iteration 2](https://github.com/Kotlin/kotlinx-rpc/pull/696/checks) — all passed |

### Failed

None.

Iteration 1 (`cd48fcd0`'s parent `1f0fbf0c`) failed on Apple with `ClientInterceptorTest.throw during intercept - should fail with thrown exception[macosArm64]` hitting the same "Test running process exited unexpectedly" pattern. Root cause was a path the initial fix missed (interceptor throws before `start()`, so `tryToCloseCall` never fires). Fixed in iteration 2 (see previous comment). Both TC builds and all GH Actions pass on iteration 2.

PR #725, opened 2026-06-11
[protoc-gen] Add an option to generate lowerCamelCase gRPC method names

### LINKED ISSUE #722: [protoc-gen] Naming convention for generated method names
Thanks for all the work on kotlinx.rpc!

Currently, `protoc-gen` generates Kotlin rpc methods using the same casing as the rpc names defined in the `.proto` file.

For example:
`sample.proto`
```proto
service ExampleService {
  rpc ExampleMethod(PingRequest) returns (PingResponse);
}

message PingRequest {}

message PingResponse {}
```

Generated kotlin code:
```kotlin
@Grpc
interface ExampleService {
    suspend fun ExampleMethod(message: PingRequest): PingResponse
}
```

As a result, the generated Kotlin method does not follow the [Kotlin naming convention of function names](https://kotlinlang.org/docs/coding-conventions.html#function-names), which recommends `lowerCamelCase`.

Although Protocol Buffers do not require any particular casing for rpc names, many examples and style guides use `PascalCase`(`TitleCase`) rpc names.

- [Protocol Buffers Documentation - Style guide](https://protobuf.dev/programming-guides/style/)
- [Buf Docs - best practice - style guide](https://buf.build/docs/best-practices/style-guide/)

Existing libraries such as [grpc-java](https://github.com/grpc/grpc-java/tree/master/compiler) and [grpc-kotlin](https://github.com/grpc/grpc-kotlin) also convert rpc names to `lowerCamelCase` when generating.

Would it make sense for the Kotlin generator to convert rpc names to `lowerCamelCase` as well?
If this direction aligns with the project's goals, I'd be happy to contribute an implementation.

hondaya14:
This PR implements https://github.com/Kotlin/kotlinx-rpc/issues/722 .

It adds the `useLowerCamelCaseGrpcMethodNames` Gradle option, which controls whether generated gRPC method names are converted to lower camel case.


protoc-gen/grpc/src/main/kotlin/kotlinx/rpc/protoc/gen/grpc/ModelToGrpcKotlinCommonGenerator.kt
```diff
@@ -52,26 +53,38 @@ class ModelToGrpcKotlinCommonGenerator(
...
-                val annotations = when (method.dec.options.idempotencyLevel) {
-                    null, DescriptorProtos.MethodOptions.IdempotencyLevel.IDEMPOTENCY_UNKNOWN -> {
-                        emptyList()
-                    }
 
-                    DescriptorProtos.MethodOptions.IdempotencyLevel.IDEMPOTENT -> {
-                        listOf("@%T(idempotent = true)".scoped(FqName.Annotations.GrpcMethod))
-                    }
+                val methodName = if (config.useLowerCamelCaseGrpcMethodNames) {
+                    KotlinKeywords.escapeIfKeyword(name = method.name.replaceFirstChar { it.lowercase() })
```
Mr3zee: This is not a CamelCase conversion, just lower the first char. `_` symbols will stay the same

gradle-plugin/src/main/kotlin/kotlinx/rpc/buf/BufExtensions.kt
```diff
@@ -203,6 +203,14 @@ public open class BufGenerateExtension @Inject internal constructor(internal val
      */
     public val indentSize: Property<Int> = project.objects.property<Int>().convention(4)
 
+    /**
+     * Whether to generate Kotlin gRPC method names in lower camel case.
+     *
+     * Default value: `true`.
+     */
+    public val useLowerCamelCaseGrpcMethodNames: Property<Boolean> =
```
Mr3zee: Sorry for changing my mind about the naming, but let's call it `camelCaseGrpcMethods`. Should be descriptive enough in the context.

protoc-gen/codegen.md
```diff
@@ -267,23 +270,28 @@ service StreamingTestService {
...
-Reads `MethodOptions.IdempotencyLevel` from the `.proto` method options and maps to `@GrpcMethod` annotation parameters:
+Reads `MethodOptions.IdempotencyLevel` from the `.proto` method options and maps to `@Grpc.Method` annotation parameters:
 
 | Proto `idempotency_level` | Generated annotation |
 |---|---|
 | (default) | _(none)_ |
-| `IDEMPOTENT` | `@GrpcMethod(idempotent = true)` |
-| `NO_SIDE_EFFECTS` | `@GrpcMethod(idempotent = true, safe = true)` |
+| `IDEMPOTENT` | `@Grpc.Method(idempotent = true)` |
+| `NO_SIDE_EFFECTS` | `@Grpc.Method(idempotent = true, safe = true)` |
```
Mr3zee, line 294: Thanks for fixing that!

Mr3zee (changes requested):
Thank you for your contribution and sorry for a delay in review. Please, see the comments below. Please, run `updateLegacyAbi` and `detekt` in the gradle plugin. Also, please rebase onto the recent version of main

hondaya14:
I ran `updateLegacyAbi` in d92be54acb57d344bec06c0bc07eeb764771bb52 .
I also ran `../gradlew -p gradle-plugin detekt`, and it reported `MaxLineLength` errors in `gradle-plugin/src/main/kotlin/kotlinx/rpc/protoc/ProtoSourceSet.kt`, 
However, these errors seem to be outside the scope of this pr. They also seem to be related to the following pr.
- https://github.com/Kotlin/kotlinx-rpc/pull/679

Could you let me know whether I can ignore these errors for this pr, or whether I should fix them here?

gradle-plugin/src/main/kotlin/kotlinx/rpc/buf/BufExtensions.kt
hondaya14: That makes sense. Thanks for the suggestion!
I fixed it in d7b497785f785fb0d8568f5bd167d4ab644320a3.

protoc-gen/grpc/src/main/kotlin/kotlinx/rpc/protoc/gen/grpc/ModelToGrpcKotlinCommonGenerator.kt
hondaya14: If I understand correctly, the method name is already converted from snake_case to camelCase by `simpleProtoNameToKotlin()` when creating the model (`MethodDeclaration`) from `MethodDescriptor`.

ref:
https://github.com/Kotlin/kotlinx-rpc/blob/5e250a24ae55f8ccb57db0735a79808a60c109da/protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/codeRequestToModel.kt#L498-L508
https://github.com/Kotlin/kotlinx-rpc/blob/5e250a24ae55f8ccb57db0735a79808a60c109da/protoc-gen/common/src/main/kotlin/kotlinx/rpc/protoc/gen/core/codeRequestToModel.kt#L605-L621

In particular, snake_case names should already be onverted to Kotlin-style method names at that stage. Given that, I think the remaining transformation needed in this generator code path. Does that sound correct?

I also added some unit tests in commit 05a5eed02a97717a4f5a94bb12bcbfaa53fb7d8b , since I think the generator behavior is hard to verify from the implementation alone.
Please let me know if there are any issues with either the expected generator behavior or the test implementation.

Mr3zee: Oh, you are correct, apologies. 
And you found a bug actually! This is not correct behavior, grpc methods like `some_method` won't work right now. 

Do you want to fix that too? If so, this is the lead:

We need to preserve the original proto name and use it unless your option is turned on. And for protobuf names for consistency. So the option can be `camelCaseNames` which controls both:
- grpc methods and message properties are lower case
- interface and class names are upper case

Implicitly this should be intuitive enough, so we don't need to put it into the name of the config parameter.

Please tell me if you want to take it, otherwise I can do that too.

Mr3zee:
@hondaya14 no, please ignore the ones not related to your changes

protoc-gen/grpc/src/main/kotlin/kotlinx/rpc/protoc/gen/grpc/ModelToGrpcKotlinCommonGenerator.kt
hondaya14: Thanks for the clarification!

I'd like to clarify the intended behavior, especially the role of the new option.
Currently, `proto: rpc some_method` is already generated as `Kotlin: fun someMethod`, while `proto: rpc SomeMethod` stays `Kotlin: fun SomeMethod`.
My understanding of the desired naming rule is that generated Kotlin APIs should follow Kotlin coding conventions, so RPC methods should be generated as lowerCamelCase Kotlin functions.

For example:
- `some_method` -> `someMethod`
- `SomeMethod` -> `someMethod`

So I think we agree on the conversion rule when camel-casing is applied.

What I’d like to clarify is the behavior when the new option is disabled. 
I thought the option was mainly needed to preserve the current generated API behavior for backward compatibility. 
However, if we use the original proto name when the new option is disabled, then `some_method` would be generated as `some_method`, which is different from the current behavior. Is that the intended behavior?

I'm not sure we need an option specifically for preserving the original proto name in the generated Kotlin API.

Mr3zee: > However, if we use the original proto name when the new option is disabled, then some_method would be generated as some_method, which is different from the current behavior. Is that the intended behavior?

Yes, you are correct, and this is intended. While indeed we will break some people's code, this is a dev build version where this is allowed. And the sooner we do it, the better. As previous behavior was not correct, breaking it is ok

hondaya14: I see. That makes sense.
I’ll update the implementation accordingly.

hondaya14: Hi @Mr3zee , sorry for the delay.

I’ve updated the implementation. The existing tests still need to be updated to match the new code generation behavior, but those changes are likely to be large.

Before I update the tests, could you review the implementation approach?

Mr3zee: Hi @hondaya14 ! Yes, that looks exactly the right way! Please also add tests for proto -> camelCase name generation with weird amount of underscores, digits, etc.

hondaya14: Hi @Mr3zee ! I've updated existing tests and added the requested test cases with multiple underscores and digits. Could you take another look?

Mr3zee (changes requested):
Hi @hondaya14 ! Thank you for this update! I'm running tests now.

All is good, expect the snake -> camel case convertion rules. Let's adopt these:
- Remove all underscores (_), except for leading and trailing
- Treat any amount of underscores in a row as one, except for leading and trailing
- Keep the first word entirely lowercase.
- Capitalize the first letter of every subsequent word.
- Join the words together with no spaces.

Mr3zee:
Please also fix the gradle tests (if unrelated, rebase on main, should be fixed there)

hondaya14:
Hi @Mr3zee,
Thank you for your review! I've updated the implementation based on your feedback.

> Keep the first word entirely lowercase.

However, implementing this rule correctly seems to require several additional considerations.
For example, converting the following names into the following results would require detecting word boundaries, including acronym boundaries.
- `GetUser_method` -> `getUserMethod`
- `userID_info` -> `userIdInfo`
- `HTTPServer_status` -> `httpServerStatus`

Supporting these cases may make the implementation more complex, since each segment would need to be further analyzed and normalized.
Could you please confirm whether this level of normalization is actually intended?

For reference, grpc-kotlin does not appear to perform this kind of acronym-aware normalization either. ref: https://github.com/grpc/grpc-kotlin/blob/master/compiler/src/main/java/io/grpc/kotlin/generator/protoc/ProtoMethodName.kt

Mr3zee:
> Could you please confirm whether this level of normalization is actually intended?

No, this is hardcore. Simple rules are ok, like "Keep the first letter lowercase.'

But I expect this work tbh, because accidentally it just obeys our rules, doesn't it?
- GetUser_method -> getUserMethod

Mr3zee:
All lgmt, I'll run the tests

Mr3zee (approved):
Lgtm, thanks for your contribution and long work on the PR!

PR #727, opened 2026-06-17
Fix compiler plugin compatibility with Kotlin 2.4.20

eminasljivic:
**Subsystem**
Compiler Plugin

**Problem Description**
`IrAnnotation` was separated from `IrConstructorCall` and the extension method `IrConstructorCall.getValueArgument` was replaces by a direct call to `argumentMapping`.

**Solution**
From 2.4.20 onwards `argumentMapping` is called directly instead of using the extension method.


Mr3zee (approved):
Very nice!
Please trigger builds for all versions in TeamCity before merging!

PR #728, opened 2026-06-23
KRPC-560: Fix Native segfault in callables property itable dispatch (…

eminasljivic:
**Subsystem**
Compiler Plugin

**Problem Description**
Native build fails due to public getter on private property.

**Solution**
Cherry picking solution from main branch (KRPC-560).


Mr3zee (approved):
LGTM!

Mr3zee:
@eminasljivic please run relevant TC builds before merging

PR #734, opened 2026-06-29
Fix logger memory leak from per-instance logger names

### LINKED ISSUE #541: Memory leak caused by creating a unique logger for each `KrpcServerService` instance
krpc version: 0.10.1

I found that `kotlinx.rpc.krpc.server.internal.KrpcServerService` creates a unique logger for each instance:
```kotlin
private val logger = RpcInternalCommonLogger.logger(rpcInternalObjectId(descriptor.fqName))
```

Since `KrpcServerService` is created on each service registration (basically, on each new connection) over time this leads to a huge number of loggers created and kept in memory by logging framework implementations (e.g. Logback).

A simple example reproducing the issue: https://github.com/akowal/krpc-log-memleak

Here's a couple of heap charts:

Image

Image

And some picks from the heapdump after JVM OOM (can be found in the example repo):

Image

Image

nickjn92:
## What

Fix a memory leak where kRPC loggers accumulated without bound on long-running servers, by switching to stable per-type logger names. Fixes #541.

## Why

KrpcServerService, KrpcServer and KrpcClient derived their logger names from rpcInternalObjectId(), which embeds the instance hashCode() in hex, e.g. KrpcServerService[com.example.MyService][3f2a1b]. A KrpcServerService is created per service registration, i.e. per connection, and logging backends such as Logback cache one logger per name and never evict. Every connection therefore left a logger behind for the lifetime of the process.

Beyond the leak, a per-instance id doesn't belong in the logger name. Logger names should form a stable, bounded set that users can configure levels and filters against; embedding identity defeats that, and a hashCode isn't collision-free, so it wasn't a reliable id to begin with. If telling instances apart in logs is ever needed, that information belongs in the logging context (e.g. MDC) or in the log message itself.

## The fix:

- KrpcClient / KrpcServer name their loggers from this::class.
- KrpcServerService uses KrpcServerService[<service fqName>], keeping per-service granularity without the per-instance part.
- rpcInternalObjectId() and ObjectId.kt are removed. There are no other call sites, and the helper is excluded from the public ABI (@InternalRpcApi, kotlinx.rpc.krpc.internal package), so no dump changes.

## How to verify

- New LoggerLeakTest serves a unary call over 64 independent client/server pairs and asserts exactly one cached logger per type remains in the Logback LoggerContext. Red on the old per-instance names (64 loggers), green after the fix.
- Full :krpc:krpc-test:jvmTest suite green: 164 tests, 0 failures.
- checkLegacyAbi passes for all published modules; no .api / .klib.api dumps changed.

---
Co-authored with Claude (via Claude Code).


Mr3zee:
Hi @nickjn92 ! Thanks for the contribution, we will review it by the end of the week!

Mr3zee (approved):

Mr3zee:
Thanks for the contribution!

PR #739, opened 2026-07-01
KRPC-585: Define BSR in Gradle plugin

### LINKED ISSUE #694: [gRPC] Cannot find a way to add dependencies to buf
**Version 0.11.0-grpc-187**

**Is your feature request related to a problem? Please describe.**
I can't find a way to tell the kotlinx-rpc plugin about BSR deps for example buf.build/googleapis/googleapis

**Describe the solution you'd like**

Something shaped like this seems natural but doesn't exist:

```
rpc {
    protoc {
        buf {
            generate {
                includeImports = true
                includeWkt = true
            }
            // no place to declare `buf.build/googleapis/googleapis`
        }
    }
 }
```

**Describe alternatives you've considered**
Two workarounds I've actually used while migrating from Wire:

1. **Patch the generated `buf.yaml` in a `doFirst` hook.**
2. **Copy the BSR protos into `src/<sourceSet>/proto/`.**

eminasljivic:
**Subsystem**
Gradle Plugin, Buf

**Problem Description**
Fixes #694
YT: [KRPC-585](https://youtrack.jetbrains.com/issue/KRPC-585): Currently there is no possibility to define BSR modules so that they can be imported in a proto files. 

**Solution**
Add `buf.deps` property with `.module(<module_name>)` and `lockFile`. If modules were defined then the new task `BufLockTask` will run and either generate a `buf.lock` file with `buf dep update` or copy the provided lock file.

In addition instead of using `workingDir` as an input the three generated files `buf.yaml`, `buf.gen.yaml` and `buf.lock` are now marked as input. This is due to the fact that `buf dep update` generates files in the directory and would cause the `BufGenerateTask` to run every time even though the relevant files didn't change.


Mr3zee:
Hey, nice work!

Two things I notice before I do an actual review:

- Please add free BSR account token to the CI, otherwise we might run out of free no account quota
- I see that you can't add deps per source set. Please address that and add comprehensive tests (deps should be transitive between source sets, additive in android variants case)

Mr3zee:
Hi @eminasljivic ! I see unrelated commits in the branch, can you please remove them?

gradle-plugin/src/main/kotlin/kotlinx/rpc/buf/BufExtensions.kt
```diff
@@ -256,3 +274,33 @@ public open class BufCommentsExtension @Inject internal constructor(internal val
...
 }
+
+public class BsrModule(
+    public val name: String,
+    public var version: String? = null
+) {
+    public infix fun version(version: String) {
+        this.version = version
+    }
+}
```
Mr3zee: This seems to be unnecessary. It looks similar to the `plugins` block, but this is the only similarity, and usually Gradle deps have a version either in the version catalog or in the string itself.

gradle-plugin/src/main/kotlin/kotlinx/rpc/buf/BufExtensions.kt
```diff
@@ -256,3 +274,33 @@ public open class BufCommentsExtension @Inject internal constructor(internal val
...
+     * BSR modules defined by their name and optionally followed by a colon and either a commit id or tag.
+     */
+    public val modules: ListProperty<BsrModule> = project.objects.listProperty<BsrModule>()
+
+    /**
+     * BSR modules defined by their name and optionally followed by a colon and either a commit id or tag.
+     */
+    public fun module(name: String): BsrModule {
+        return BsrModule(name).also { modules.add(it) }
+    }
```
Mr3zee: Can you please add an override for the version catalog defined deps, and a test case for it?

gradle-plugin/src/main/kotlin/kotlinx/rpc/protoc/ProtocBufDeps.kt
```diff
@@ -0,0 +1,29 @@
...
+    project: Project,
+    sourceSetName: String
+) : BufDepsExtension(project) {
+    /**
+     * Path to the buf.lock file. Used as input if it exists, or generated here if it doesn't.
+     * Default: buf/<sourceSet>/buf.lock.
+     */
+    public val lockFile: Property<String> = project.objects.property<String>().convention(
+        "buf/$sourceSetName/buf.lock"
+    )
```
Mr3zee: We can put this into the BufDepsExtension itself without a separate class.

Mr3zee (changes requested):
Lgtm! Few small comments and we can merge

gradle-plugin/src/main/kotlin/kotlinx/rpc/buf/BufExtensions.kt
eminasljivic: It's not possible to define a buf module in the version catalog, since it only supports the 'group:artifact:version' coordinates. It's possible to define it within the version block but that is only a string and that would work with what we already have.

Mr3zee: Oh, sad
ok, than all done

Mr3zee (approved):

PR #742, opened 2026-07-01
gRPC: Add channel-level userAgent configuration to client

mlilienberg:
# PR 2 — channel-level userAgent

- **Title:** Add channel-level userAgent configuration to gRPC client
- **Base:** `Kotlin/kotlinx-rpc:main`  ←  **Head:** `mlilienberg:feat/grpc-client-user-agent-pr`
- **Open with:** `gh pr create -R Kotlin/kotlinx-rpc --base main --head mlilienberg:feat/grpc-client-user-agent-pr --title "Add channel-level userAgent configuration to gRPC client" --body-file PR_DRAFT_grpc-client-user-agent.md`

---

## Summary

Adds a `userAgent` option to `GrpcClientConfiguration`, applied at the channel/transport
level so it is actually sent on the wire. A `user-agent` entry added to request
`GrpcMetadata` is a reserved header and is dropped/overwritten by the gRPC runtime, so this
is the correct way to influence the `User-Agent`.

The value is **prepended**, not set: the runtime always appends its own token after a space,
so the wire value is `"<userAgent> <runtime-token>"`. This matches the behavior of the other
gRPC client implementations (grpc-go's `WithUserAgent`, grpc-java's `ManagedChannelBuilder.userAgent`,
grpc-swift's `userAgentSuffix`/prefix handling), which all treat the application user-agent as a
prefix to the runtime's own token rather than a full replacement. Suppressing the appended token
(a full override) is not supported by the underlying gRPC implementations.

For example, with:

```kotlin
GrpcClient("example.com", 443) {
    userAgent = "MyApp/1.2.3"
}
```

the `User-Agent` actually sent on the wire is:

- JVM: `MyApp/1.2.3 grpc-java-okhttp/<version>` (or `grpc-java-netty/<version>`)
- Native: `MyApp/1.2.3 grpc-c/<version> (<platform>; <engine>)`

## Motivation

`GrpcClientConfiguration` already mirrors common channel-level knobs from the other gRPC
implementations (keep-alive, override-authority, interceptors), but there was no way to set the
application `User-Agent`. This aligns `kotlinx.rpc`'s gRPC client with grpc-go, grpc-java,
grpc-kotlin, and grpc-swift, all of which expose a channel-level user-agent prefix — letting
servers and proxies identify the calling application (version, platform) in logs and routing.

## Changes

- `GrpcClientConfiguration.userAgent: String?` (default `null`), fully documented.
- JVM: maps to `ManagedChannelBuilder.userAgent(...)`.
- Native: maps to the `grpc.primary_user_agent` (`GRPC_ARG_PRIMARY_USER_AGENT_STRING`) channel arg.
- Regenerated ABI dumps (`grpc-client.api` / `grpc-client.klib.api`).
- Tests (`GrpcUserAgentTest`) mirroring the existing `GrpcKeepAliveTest` pattern.

## Verification

- `./gradlew :grpc:grpc-core:jvmTest --tests "*GrpcUserAgentTest*"` — passes (asserts the composed
  channel user-agent starts with the configured prefix).
- `./gradlew :grpc:grpc-core:macosArm64Test --tests "*GrpcUserAgentTest*"` — passes (asserts the
  `grpc.primary_user_agent` channel arg from trace logs equals the configured value).
- `./gradlew :grpc:grpc-client:detekt :grpc:grpc-core:detekt` — passes.

## AI use disclosure

Co-authored with Claude (implementation review, tests, ABI regeneration, verification).

### DIFF
--- a/grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/GrpcClient.kt
+++ b/grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/GrpcClient.kt
@@ -261,8 +261,7 @@ public class GrpcClientConfiguration internal constructor() {
      * Set this here rather than via request [kotlinx.rpc.grpc.GrpcMetadata]: `user-agent` is a reserved
      * header and any value added to the metadata is overwritten by the gRPC runtime.
      *
-     * Maps to `ManagedChannelBuilder.userAgent(...)` on JVM and the `GRPC_ARG_PRIMARY_USER_AGENT_STRING`
-     * channel argument on native. If `null` (the default), only the runtime's own token is sent.
+     * If `null` (the default), only the runtime's own token is sent.
      *
      * ```
      * GrpcClient("example.com", 443) {
--- a/grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/GrpcClient.kt
+++ b/grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/GrpcClient.kt
@@ -250,6 +250,27 @@ public class GrpcClientConfiguration internal constructor() {
      */
     public var overrideAuthority: String? = null
 
+    /**
+     * A custom application `User-Agent` for this channel.
+     *
+     * The value is used as a **prefix**: the runtime appends its own token after a single space, so the
+     * `User-Agent` on the wire is `"<userAgent> <runtime-token>"` (e.g. `grpc-java-okhttp/<version>` on
+     * JVM, or the `grpc-c/<version>` C-core token on native). A full override is not supported by the
+     * underlying gRPC implementations.
+     *
+     * Set this here rather than via request [kotlinx.rpc.grpc.GrpcMetadata]: `user-agent` is a reserved
+     * header and any value added to the metadata is overwritten by the gRPC runtime.
+     *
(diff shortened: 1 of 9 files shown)


grpc/grpc-client/src/commonMain/kotlin/kotlinx/rpc/grpc/client/GrpcClient.kt
```diff
@@ -250,6 +250,27 @@ public class GrpcClientConfiguration internal constructor() {
...
+     * The value is used as a **prefix**: the runtime appends its own token after a single space, so the
+     * `User-Agent` on the wire is `"<userAgent> <runtime-token>"` (e.g. `grpc-java-okhttp/<version>` on
+     * JVM, or the `grpc-c/<version>` C-core token on native). A full override is not supported by the
+     * underlying gRPC implementations.
+     *
+     * Set this here rather than via request [kotlinx.rpc.grpc.GrpcMetadata]: `user-agent` is a reserved
+     * header and any value added to the metadata is overwritten by the gRPC runtime.
+     *
+     * Maps to `ManagedChannelBuilder.userAgent(...)` on JVM and the `GRPC_ARG_PRIMARY_USER_AGENT_STRING`
+     * channel argument on native. If `null` (the default), only the runtime's own token is sent.
```
Mr3zee: You can just leave the `If `null` (the default), only the runtime's own token is sent.` part

Mr3zee (changes requested):
LGMT, please just change the comment content

Mr3zee (approved):

PR #759, opened 2026-09-01
KXRPC-674: Ignore proto files that are already generated by an imported source set

eminasljivic:
**Subsystem**
Gradle Plugin, Protoc

**Problem Description**
[KXRPC-674](https://youtrack.jetbrains.com/issue/KXRPC-674): Imported proto messages duplicated in dependent source sets with includeImports

The problem is that when "includeImports" is set it could cause that some proto files get generated multiple times.

**Solution**

The protoc plugins write the `fileToGenerateList` in the sourceSets folder into the folder `protoInputFiles`. If the plugin is run using the strategy `directory` and the input files are nested within directories, the plugin would run for each directory separately. Thus the `protoInputFiles` follow the directory structure from the input files.

<details>

<summary>Example</summary>

A project with this proto file structure:
proto/some.proto
proto/nested/nested.proto

The plugin would run twice. For the main folder the `fileToGenerateList` would contain only `some.proto` and thus it would generate only `protocInputFiles/some.proto.txt`. The same thing with the respective file for the `nested` folder.

The content of the files is only the proto file name.

</details>

The GenerateBufGenYaml task then reads these files of all the imported source sets and sets the new option "ingnoreFiles" for the plugins. The plugin then filter the `fileToGenerateList` using this new option. 

This means that now the GenerateBufGenYaml depends on the GenerateBuf task, since only then are the plugins run for the imported source sets.

Old task tree:

```
extractProtoCommonMain  ──────┐ 
                              ├─> processProtoCommonMain ─> processImportProtoCommonMain ─> generateBufYamlCommonMain ─> generateBufGenYamlCommonMain ─> bufLockCommonMain ─> bufGenerateCommonMain
extractProtoImportCommonMain ─┘                                                                                                                                                           │
                                                                                                                                                                                          │ 
                       extractProtoJvmMain  ──────┐                                                                                                                                       ▼
                                                  ├─> processProtoJvmMain ─> processImportProtoJvmMain ─> generateBufYamlJvmMain ─> generateBufGenYamlJvmMain─> bufLockJvmMain ─> bufGenerateJvmMain
                       extractProtoImportJvmMain ─┘
```

New task tree: 

```
extractProtoCommonMain  ──────┐ 
                              ├─> processProtoCommonMain ─> processImportProtoCommonMain ─> generateBufYamlCommonMain ─> generateBufGenYamlCommonMain ─> bufLockCommonMain ─> bufGenerateCommonMain
extractProtoImportCommonMain ─┘                                                                                                                                                        │
                                                                                                                                                                                       │ 
                                                              extractProtoJvmMain  ──────┐                                                                                             ▼
                                                                                         ├─> processProtoJvmMain ─> processImportProtoJvmMain ─> generateBufYamlJvmMain ─> generateBufGenYamlJvmMain ─> bufLockJvmMain ─> bufGenerateJvmMain
                                                              extractProtoImportJvmMain ─┘
```


gradle-plugin/src/main/kotlin/kotlinx/rpc/buf/tasks/GenerateBufGenYaml.kt
```diff
@@ -95,11 +102,23 @@ public abstract class GenerateBufGenYaml @Inject internal constructor(
...
                 }
 
+                val ignoreFiles = importedProtocIgnoreFileDirectories.get()
+                    .flatMap { it.walk().toList() }
+                    .filter { it.isFile }
+                    .flatMap { it.readLines() }
+                    .distinct()
+
+                val options = plugin.options + if (ignoreFiles.isNotEmpty()) {
+                    mapOf("ignoreFiles" to ignoreFiles.joinToString(";"))
```
Mr3zee: Suprisingly, ';' can be a valid path name in many OSes, so it's better to use null-byte sparator `\0` or JSON array here

Mr3zee (approved):
Very nice work, thank you! Only have one nitpick

Jozott00 (approved):
I don't see any issues, LGTM

